PR #13, opened 2024-02-16
Fixes downloading large results

merschformann:
# Description

Fixes downloading files that exceed the direct request's limit.

## Changes

- Fixes large result download

### DIFF
--- a/nextmv/cloud/application.py
+++ b/nextmv/cloud/application.py
@@ -5,8 +5,6 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Any
 
-import requests
-
 from nextmv.base_model import BaseModel
 from nextmv.cloud.acceptance_test import AcceptanceTest, Metric
 from nextmv.cloud.batch_experiment import BatchExperiment, BatchExperimentMetadata, BatchExperimentRun
--- a/nextmv/cloud/application.py
+++ b/nextmv/cloud/application.py
@@ -724,8 +724,11 @@ class Application:
             return result
 
         download_url = DownloadURL.from_dict(response.json()["output"])
-        download_response = requests.get(download_url.url)
-        download_response.raise_for_status()
+        download_response = self.client.request(
+            method="GET",
+            endpoint=download_url.url,
+            headers={"Content-Type": "application/json"},
+        )
         result.output = download_response.json()
 
         return result
--- a/nextmv/cloud/application.py
+++ b/nextmv/cloud/application.py
@@ -5,6 +5,8 @@ from dataclasses import dataclass
 from datetime import datetime
 from typing import Any
 
+import requests
+
 from nextmv.base_model import BaseModel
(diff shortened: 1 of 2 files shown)


nextmv/cloud/application.py
```diff
@@ -717,13 +719,13 @@ def _run_result(
...
         if not large_output:
-            return RunResult.from_dict(response.json())
+            return result
 
-        download_url = DownloadURL.from_dict(response.json())
-        response2 = self.client.request(
-            method="GET",
-            endpoint=download_url.url,
-        )
+        download_url = DownloadURL.from_dict(response.json()["output"])
```
sebastian-quintero, line 724: What if we do `client2=copy.deepcopy(client)` and then `client2.headers = {}`? That way we can use `client2.request(...)`, the way it was?

merschformann: I thought about that, but it seemed a bit overcomplicated (for what it needs to do) and more brittle to me. After all, the pre-signed download url is a classic GET request.
But yeah, no strong opinion here. I assumed that your style may vary, so, feel free to change it. Just wanted to drop my motivation for choosing this style. :relaxed:

nextmv/cloud/application.py
```diff
@@ -717,13 +719,13 @@ def _run_result(
...
+        download_response = requests.get(download_url.url)
+        download_response.raise_for_status()
+        result.output = download_response.json()
```
sebastian-quintero:
```suggestion
        download_response = self.client.request(
            method="GET",
            endpoint=download_url.url,
            headers={"Content-Type": "application/json"},
        )
        result.output = download_response.json()
```

Actually I think this is what we should do. I introduced the `headers` argument in the `request` method in case one doesn’t want to use the client’s default headers.

nextmv/cloud/application.py
sebastian-quintero: Please see my other comment.

nextmv/cloud/application.py
merschformann: Yeah, good call. That is clean again. :relaxed:

nextmv/cloud/application.py
sebastian-quintero: It would be nice to make sure it works.

merschformann: On it. :100:

merschformann: Confirmed working. :+1:

pyproject.toml
```diff
@@ -46,7 +46,7 @@ Repository = "https://github.com/nextmv-io/nextmv-py"
 
 [tool.ruff]
 target-version = "py312"
-select = [
+lint.select = [
```
merschformann, line 49: This was a deprecation warning by ruff @sebastian-quintero

sebastian-quintero (approved):
🌮 for catching the bug @merschformann!

PR #19, opened 2024-04-02
Allows non-JSON inputs for runs

merschformann:
# Description
This update allows the `new_run` method in the Application class to accept text inputs directly, alongside the existing dict and BaseModel inputs.

# Changes
- Added string type as an acceptable input format for runs.
- Modified input size calculation and upload logic to accommodate the new input type.
- Added `upload_to_presigned_url` method to handle upload put requests.

### DIFF
--- a/nextmv/cloud/client.py
+++ b/nextmv/cloud/client.py
@@ -3,7 +3,7 @@
 import json
 import os
 from dataclasses import dataclass, field
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional
 from urllib.parse import urljoin
 
 import requests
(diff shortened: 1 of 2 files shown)


larsbeck (approved):

sebastian-quintero (approved):
LGTM

nextmv/cloud/client.py
```diff
@@ -180,6 +180,39 @@ def request(
 
         return response
 
+    def upload_to_presigned_url(
```
sebastian-quintero: Why do we need this new method? Can't we use the one we already have? What is the main issue tht would prevent us from using it?

On another note, this method doesn't use "self", should it really be a class method?

One drawback of not using the other method is that we have no timeouts, retries, backoff, or anything.

sebastian-quintero (commented):
I would like some clarity on why another method is needed! Thanks.

nextmv/cloud/client.py
merschformann: We talked about this offline and will aim to keep it with the existing logic as much as possible.

PR #20, opened 2024-04-03
Fix signatures and required info in batch

sebastian-quintero:
# Description

Modifies some signatures used in batch experiments.

- Option sets are more succint.
- Instance ids are optional when starting a batch experiment.
- In a run, either a version or an instance must be specified.


ryanjoneil (approved):
👏

PR #21, opened 2024-04-08
ENG-4772 Handle non-existent batch exp and change arg names

sebastian-quintero:
# Description

Handle the error when the `input_set_id` is not provided and the batch experiment does not exist. Change the arg names so that it matches the CLI.


merschformann (approved):
:tada:

PR #22, opened 2024-04-08
Support for StatusV2

sebastian-quintero:
# Description

Adds support for `status_v2`. For the polling logic, uses the additional `canceled` state as criteria for stopping the polling action. Additionally, new methods are introduced to cancel runs and list the account queue, which potentially may be done programatically.

### DIFF
--- a/nextmv/cloud/account.py
+++ b/nextmv/cloud/account.py
@@ -1,0 +1,69 @@
+from dataclasses import dataclass
+from datetime import datetime
+from typing import List
+
+from nextmv.base_model import BaseModel
+from nextmv.cloud.client import Client
+from nextmv.cloud.status import Status, StatusV2
+
+
+class QueuedRun(BaseModel):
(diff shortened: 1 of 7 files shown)


merschformann (approved):
❤️

merschformann (approved):

nextmv/cloud/account.py
```diff
@@ -0,0 +1,69 @@
...
+    execution_class: str
+    """Execution class used for the run."""
+    status: Status
+    """Deprecated: use status_v2."""
+    status_v2: StatusV2
+    """Status of the run."""
+
+
+class Queue(BaseModel):
+    """A queue is a list of runs that are pending to be executed in the
```
mamclaughlin: pending to be executed or currently being executed

sebastian-quintero: Listo!

PR #23, opened 2024-04-21
Adds support for retrieving run logs

merschformann:
# Description

Adds support for retrieving logs of a run.

## Changes

- Adds `run_logs` func to the application for retrieving logs of a run

### DIFF
--- a/nextmv/cloud/application.py
+++ b/nextmv/cloud/application.py
@@ -103,12 +103,19 @@ class RunInformation(BaseModel):
 
 
 class RunResult(RunInformation):
-    """Result of a run, wheter it was successful or not."""
+    """Result of a run, whether it was successful or not."""
 
     error_log: Optional[ErrorLog] = None
     """Error log of the run. Only available if the run failed."""
     output: Optional[Dict[str, Any]] = None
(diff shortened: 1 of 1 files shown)


sebastian-quintero (approved):
Non-blocking remark.

nextmv/cloud/application.py
```diff
@@ -646,6 +653,28 @@ def run_input(
...
+            Logs of the run.
+
+        Raises:
+            requests.HTTPError: If the response status code is not 2xx.
+        """
+        response = self.client.request(
+            method="GET",
+            endpoint=f"{self.endpoint}/runs/{run_id}/logs",
+        )
+        return RunLog.from_dict(response.json())
```
sebastian-quintero, line 676: I think the pattern we are following so far is to return the actual class. People can then do whatever they want with it, like converting it to a dict. Up to you.

sebastian-quintero (approved):
The comment didn’t come through. Minor remark.

nextmv/cloud/application.py
merschformann: Good catch! Fixed it. :hugs:

PR #25, opened 2024-05-10
ENG-5031 Adds method for deleting an acceptance test

sebastian-quintero:
# Description

Adds a method for deleting an acceptance test.

### DIFF
--- a/nextmv/cloud/application.py
+++ b/nextmv/cloud/application.py
@@ -214,15 +214,13 @@ class Application:
             endpoint=f"{self.endpoint}/runs/{run_id}/cancel",
         )
 
-    def delete_batch_experiment(
-        self,
-        id: str,
-    ) -> None:
+    def delete_batch_experiment(self, batch_id: str) -> None:
         """
-        Deletes a batch experiment, along with all of its runs.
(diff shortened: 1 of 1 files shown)


dirkschumacher (approved):

nextmv/cloud/application.py
```diff
@@ -631,10 +646,7 @@ def new_run_with_result(
             polling_options=polling_options,
         )
 
-    def run_input(
-        self,
-        run_id: str,
-    ) -> Dict[str, Any]:
+    def run_input(self, run_id: str) -> Dict[str, Any]:
```
merschformann, line 649: Our formatters seem to be configured differently. Or, was that a deliberate choice? :thinking:

merschformann (approved):

nextmv/cloud/application.py
sebastian-quintero: They should be configured the same, it was just a deliberate choice of the line being so short that each param doesn’t need to go on  a new line.

merschformann: Makes sense. :blush:

PR #26, opened 2024-08-22
ENG-5405 Adds support for logging

sebastian-quintero:
# Description

Adds functionality for logging. For the Nextmv platform, logging implies streaming to `stderr`.

Two new features are introduced:
- `log`: it is a function to simply print to `stderr`.
- `redirect_stdout`: to redirect stdout to stderr. This is needed for solvers that only log to stdout.
- `reset_stdout`: reset the behavior of stdout.

## Usage

Consider the following Python script:

```python
import sys

import nextmv

print("0. I do nothing")

nextmv.redirect_stdout()

nextmv.log("1. I log a message to stderr")

print("2. I print a message to stdout")

nextmv.reset_stdout()

print("3. I print another message to stdout")

print("4. I print yet another message to stderr without the logger", file=sys.stderr)

nextmv.log("5. I log a message to stderr using the nextmv module directly")

print("6. I print a message to stdout, again")

```

After executing it, here are the messages printed to the different streams.

- `stdout`

```txt
0. I do nothing
3. I print another message to stdout
6. I print a message to stdout, again
```

- `stderr`

```txt
1. I log a message to stderr
2. I print a message to stdout
4. I print yet another message to stderr without the logger
5. I log a message to stderr using the nextmv module directly
```

### DIFF
--- a/nextmv/logger.py
+++ b/nextmv/logger.py
@@ -12,16 +12,34 @@ class Logger:
     all the logs stored to stderr, and the behavior of stdout and stderr will
     be restored to their original state.
 
-    If you only want to log messages to stderr, and not redirect the stdout
-    stream to stderr, you can instead use the `log` function from this same
-    module.
+    If you only want to log messages to stderr, without buffering them, and not
+    redirect the stdout stream to stderr, you can instead use the `log`
+    function from this same module.
+
+    Examples:
+
+        - Print message "0" to stdout, messages "1" and "2" to stderr, and then "3" to stdout:
+
+        ```python
+        import nextmv
+
+        print("0. I do nothing")
+        logger = nextmv.Logger()
+        logger.log("1. I log a message to stderr")
+        print("2. I print a message to stdout, but gets redirected to stderr")
+        logger.flush()
+        print("3. I print another message to stdout")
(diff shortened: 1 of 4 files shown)


nextmv/logger.py
```diff
@@ -0,0 +1,91 @@
...
+        """
+        Flush the buffer to stderr and clear the buffer.
+        """
+
+        for log in self._logs:
+            self._original_stderr.write(log + "\n")
+
+        self._logs = []
+        sys.stdout = self._original_stdout
+        sys.stderr = self._original_stderr
```
merschformann: Only flushing on explicit call by default seems risky, as logs will be lost when exceptions occur. Should we make immediate flushing the default and only reduce flushing by request for performance reasons?

merschformann: Also, I think some kind of buffer implementation would be useful to not require the user to remember to flush. :thinking:

sebastian-quintero: Changed the whole thing.

sebastian-quintero:
Added a little something to have an "exported" version that one can reference from outside.

merschformann (approved):
:tada:

PR #27, opened 2024-08-23
ENG-5400 Adds configuration

sebastian-quintero:
# Description

Adds new `Configuration` and `Parameter` classes used for options/configurations. A parameter dictates an option that, as the bare minimum, has a name and a type. A configuration can be instantiated from a list of parameters. Parameters can be populated from command-line arguments, environment variables, or default values. A special help is provided when invoking the `-h/--help` flag on a script that uses the `Configuration`.

The `Configuration` is a namespace that allows someone to access parameters directly with the name, i.e.: `config.param_a`.

## Examples

The following examples use this python script:

```python
import nextmv

config = nextmv.Configuration(
    nextmv.Parameter("duration", str, "30s", description="solver duration", required=True),
    nextmv.Parameter("threads", int, 4, description="computer threads", required=True),
)

print(config.duration, config.threads)
```

-  Default values are used.

```bash
$ python main.py
30s 4
```

-  The `duration` parameter is taken from a command-line argument

```bash
$ python main.py -duration 40s
40s 4
```

-  The `duration` parameter is taken from an environment variable

```bash
$ DURATION=50s python main.py
50s 4
```

-  The `duration` parameter is taken from a command-line argument (highest precedence_, even though the env var is specified.

```bash
$ DURATION=50s python main.py -duration 40s
40s 4
```

- This is the help manu.

```bash
$ python main.py -h
usage: main.py [options]

Configuration for main.py. Use command-line arguments (highest precedence) or environment variables.

options:
  -h, --help            show this help message and exit
  -duration DURATION, --duration DURATION
                        [env var: DURATION] (required) (default: 30s) (type: str): solver duration
  -threads THREADS, --threads THREADS
                        [env var: THREADS] (required) (default: 4) (type: int): computer threads
```

- The bare minimum that a parameter needs is a name and a type. Consider this other python script:

```python
import nextmv

config = nextmv.Configuration(
    nextmv.Parameter("duration", str),
    nextmv.Parameter("threads", int),
)

print(config.duration, config.threads)
```

The help is a bit different:

```bash
$ python main.py -h
usage: main.py [options]

Configuration for main.py. Use command-line arguments (highest precedence) or environment variables.

options:
  -h, --help            show this help message and exit
  -duration DURATION, --duration DURATION
                        [env var: DURATION] (type: str)
  -threads THREADS, --threads THREADS
                        [env var: THREADS] (type: int)
```

But using it is straight forward:

```bash
$ python main.py -duration 30s -threads 3
30s 3
```

### DIFF
--- a/nextmv/configuration.py
+++ b/nextmv/configuration.py
@@ -1,0 +1,186 @@
+"""Configuration for a run."""
+
+import argparse
+import os
+from dataclasses import dataclass
+from typing import Any, Dict, Optional
+
+from nextmv.base_model import BaseModel
+
+
+@dataclass
+class Parameter:
+    """
+    Parameter that is used in a `Configuration`. When a parameter is required,
+    it is a good practice to provide a default value for it. This is because
+    the configuration will raise an error if a required parameter is not
+    provided through a command-line argument, an environment variable or a
+    default value.
+    """
+
+    name: str
+    """The name of the parameter."""
+    param_type: type
+    """The type of the parameter."""
+
+    default: Optional[Any] = None
+    """The default value of the parameter. Even though this is optional, it is
+    recommended to provide a default value for all parameters."""
+    description: Optional[str] = None
+    """An optional description of the parameter. This is useful for generating
+    help messages for the configuration."""
+    required: bool = False
+    """Whether the parameter is required. If a parameter is required, it will
+    be an error to not provide a value for it, either trough a command-line
+    argument, an environment variable or a default value."""
+
+
+class Configuration:
+    """
+    Configuration for a run. To initialize a configuration, pass in one or more
+    `Parameter` objects. The configuration will look for the values of the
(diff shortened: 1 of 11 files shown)


nextmv/configuration.py
```diff
@@ -0,0 +1,186 @@
...
+    description: Optional[str] = None
+    """An optional description of the parameter. This is useful for generating
+    help messages for the configuration."""
+    required: bool = False
+    """Whether the parameter is required. If a parameter is required, it will
+    be an error to not provide a value for it, either trough a command-line
+    argument, an environment variable or a default value."""
+
+
+class Configuration:
```
merschformann: Should this be called configuration? I think we use the term already for the parent of execution class. :thinking:

merschformann (approved):
Apart from the potential name clash, I like it. Should we reconsider naming?

PR #29, opened 2024-08-30
ENG-5404 Adds support for writing an output

sebastian-quintero:
# Description

Creates the `Output` and `LocalOutputWriter` classes for writing outputs to different locations.

### DIFF
--- a/nextmv/input.py
+++ b/nextmv/input.py
@@ -39,15 +39,47 @@ class Input:
         Options that the input was created with.
     """
 
-    data: Union[Dict[str, Any], str, List[Dict[str, Any]], Dict[str, List[Dict[str, Any]]]]
+    data: Union[
+        Dict[str, Any],  # JSON
+        str,  # TEXT
+        List[Dict[str, Any]],  # CSV
+        Dict[str, List[Dict[str, Any]]],  # CSV_ARCHIVE
+    ]
     """The actual data. The data can be of various types, depending on the
     input format."""
 
     input_format: Optional[InputFormat] = InputFormat.JSON
     """Format of the input data. Default is `InputFormat.JSON`."""
-
     options: Optional[Options] = None
     """Options that the `Input` were created with."""
+
+    def __post_init__(self):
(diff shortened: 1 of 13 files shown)


nextmv/input.py
```diff
@@ -108,11 +140,15 @@ def _read_json(path: str) -> Dict[str, Any]:
...
 
+    # All of these readers are callback functions.
     STDIN_READERS = {
         InputFormat.JSON: lambda: json.load(sys.stdin),
         InputFormat.TEXT: lambda: sys.stdin.read(),
         InputFormat.CSV: lambda: list(csv.DictReader(sys.stdin, quoting=csv.QUOTE_NONNUMERIC)),
     }
+    # These callbacks were not implemented with lambda because we needed
+    # multiple lines. By using `open`, we needed the `with` to be able to close
+    # the file.
```
merschformann, line 151: No shame in this. :joy_cat:

nextmv/output.py
```diff
@@ -0,0 +1,392 @@
...
+        output : Output
+            Output data to write.
+        path : str
+            Path to write the output data to.
+        """
+
+        # If the user forgot to reset stdout after redirecting it, we need to
+        # do it here to avoid unexpected behavior.
+        if stdout_redirected:
+            reset_stdout()
```
merschformann, line 352: Why do we need to reset stdout redirection here (_for all cases_)? Aren't there cases where we write to a file instead of stdout? :thinking:

sebastian-quintero: This is done so if the user wants to write to stdout, and was using our `redirect_stdout` to redirect the solver chatter to stderr, and forgot to reset stdout to its original state, we do it for them.

sebastian-quintero: Otherwise instead of printing the result to stdout they end up printing the result to stderr.

merschformann (approved):
Just some questions, don't want to block without further context.

sebastian-quintero:
> Just some questions, don't want to block without further context.

Answered your questions 😄

PR #34, opened 2024-09-06
Allow writing pure Dicts

sebastian-quintero:
- When instantiating an `Output` object that has `JSON` as the output format, check that the solution is serializable and not enforce a `dict` or `list`.
- When using the `write_local` function (really that is a convenience for `LocalOutputWriter().write`), allow the `output` to be of type `Output` or a simple `dict`. The downstream functions handle the cases when it is one way or the other.

### DIFF
--- a/nextmv/output.py
+++ b/nextmv/output.py
@@ -364,7 +364,7 @@ class LocalOutputWriter(OutputWriter):
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


merschformann (approved):
Awesome! Thank you! :hugs:

PR #37, opened 2024-09-10
Fix parsing of bool options

sebastian-quintero:
# Description

Fixes a bug where there was no possibility to obtain `False` for a bool option. Now, when a `bool` option is encountered, it is treated as a string and translated to the appropriate `bool` value.

### DIFF
--- a/nextmv/options.py
+++ b/nextmv/options.py
@@ -125,7 +125,7 @@ class Options:
             parser.add_argument(
                 f"-{param.name}",
                 f"--{param.name}",
-                type=param.param_type if param.param_type != bool else str,  # noqa E721
+                type=param.param_type if param.param_type is not bool else str,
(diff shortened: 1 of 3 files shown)


nextmv/options.py
```diff
@@ -204,3 +207,20 @@ def _description(param: Parameter) -> str:
...
+        param_type = parameter.param_type
+        if param_type is not bool:
+            return value
+
+        value = str(value).lower()
+
+        if value in ("true", "1", "t", "y", "yes"):
+            return True
+        if value in ("false", "0", "f", "n", "no"):
+            return False
```
merschformann, line 224: Nice! :heart:

merschformann (approved):
Thanks for the fast fix! :hugs:

PR #41, opened 2024-09-19
Fixes None default and dash option handling

merschformann:
# Description

Fixes `None`-default-option and dashed-option (`-`) handling.

## Changes

- Added stripping of leading '-' from parameter names to align with argparse behavior.
- Replaced '-' with '_' in parameter names for consistency with argparse and fixing bug when facing `-` options.
- Simplified default value assignment logic for non-required parameters.
- Updated tests to reflect changes in parameter requirements and default values.
- Added new tests for handling parameters with none default values and different naming conventions.

### DIFF
--- a/nextmv/options.py
+++ b/nextmv/options.py
@@ -116,11 +116,14 @@ class Options:
             description="Options for %(prog)s. Use command-line arguments (highest precedence) "
             + "or environment variables.",
         )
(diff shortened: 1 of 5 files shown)


sebastian-quintero (approved):
LGTM, I think

PR #42, opened 2024-09-21
Add options to Parameter type

ryanjoneil:
For scikit-learn and other integrations, I need to be able to do this:

```python
    nextmv.Parameter(
        "criterion",
        str,
        choices=["squared_error", "friedman_mse", "absolute_error", "poisson"],
        description="The function to measure the quality of a split.",
    )
```

This PR adds that field to the `Parameter` type. It also corrects what I believe is an error in default handling.

### DIFF
--- a/nextmv/options.py
+++ b/nextmv/options.py
@@ -52,7 +52,7 @@ class Parameter:
     """Whether the parameter is required. If a parameter is required, it will
     be an error to not provide a value for it, either trough a command-line
     argument, an environment variable or a default value."""
-    choices: List[Optional[Any]] = None
+    choices: Optional[List[Any]] = None
     """Limits values to a specific set of choices."""
 
 
--- a/nextmv/options.py
+++ b/nextmv/options.py
@@ -3,7 +3,7 @@
 import argparse
 import os
 from dataclasses import dataclass
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional
 
 from nextmv.base_model import BaseModel
 
@@ -23,6 +23,8 @@ class Parameter:
         The name of the parameter.
     param_type : type
         The type of the parameter.
+    choices : List[Any], optional
+        Limits values to a specific set of choices.
     default : Any, optional
         The default value of the parameter. Even though this is optional, it is
         recommended to provide a default value for all parameters.
@@ -50,6 +52,8 @@ class Parameter:
     """Whether the parameter is required. If a parameter is required, it will
     be an error to not provide a value for it, either trough a command-line
     argument, an environment variable or a default value."""
+    choices: List[Optional[Any]] = None
+    """Limits values to a specific set of choices."""
 
 
 class Options:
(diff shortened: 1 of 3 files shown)


nextmv/options.py
```diff
@@ -125,19 +129,25 @@ def __init__(self, *parameters: Parameter):
...
 
+            kwds = {}
+            if param.choices is not None:
+                kwds["choices"] = param.choices
             parser.add_argument(
                 f"-{param.name}",
                 f"--{param.name}",
                 type=param.param_type if param.param_type is not bool else str,
                 help=self._description(param),
+                **kwds,
```
ryanjoneil, line 140: We don't want to pass the `choices` parameter unless it is `not None`.

nextmv/options.py
```diff
@@ -163,12 +173,15 @@ def __init__(self, *parameters: Parameter):
...
-            if not param.required:
+            # Finally, attempt to set the value of a parameter from the default
+            # value.
+            if param.default is not None:
                 setattr(self, arg, param.default)
                 continue
 
+            if not param.required:
+                continue
+
```
ryanjoneil, line 184: I believe the current logic is incorrect. This is what it does today:

* If a parameter is not set and not required, the default is set.

I believe this should be:

* If a parameter is not set and has a default, then the default is set.
* If a parameter is not required and has no default, then the parameter is not set.

tests/test_options.py
```diff
@@ -362,7 +362,7 @@ def test_none_default(self):
             text=True,
         )
         self.assertEqual(result3.returncode, 0, result3.stderr)
-        self.assertEqual(result3.stdout, "{'str_opt': None}\n")
+        self.assertEqual(result3.stdout, "{}\n")
```
ryanjoneil, line 365: See the note above. Setting `default=None` will not set the parameter.

nextmv/options.py
merschformann: I just changed this behavior around to be more in line with argparse behavior. To me it seemed pretty simple, and aligns with the first case you are correctly stating. I would personally prefer it, as it is also simpler.

What is the reason to change it? I think I am missing something.

E.g., if I specify some `choices=` on argparse **and** they are _not_ required **and** they do _not_ have a default (i.e., just this: `choices=["a", "b", "c"], type=str`), I get the following when not providing the parameter: `None`. No exception is thrown on accessing the field by name (`args.choice`). 

One thing we had to fix was that providing a `None` _default_ did not work, and I think would again not work with above logic. :thinking: 

I just noticed that the test that I added unfortunately does not cause the exception like normal use would do. I opened a PR to hopefully improve the test and better explain why I have concerns (this would directly affect my use of this functionality - see #43  ). :sweat_smile:

merschformann (commented):
I like the `choices` support, but would love us to overthink changing the behavior. <3

nextmv/options.py
sebastian-quintero: From quickly reading this, I think I concur with Marius. We need to be able to set the param to `None`.

sebastian-quintero (changes requested):
I think the behavior with default and `None` doesn’t need to change, right? Because @merschformann changed it recently...

PR #45, opened 2024-09-23
ENG-5448 Nextroute moves out of nextmv-py and on to nextroute

sebastian-quintero:
# Description

All the `nextroute` stuff is moving to the `nextroute` repo as a new `nextroute` Python package will be created. The schema definitions that lived here are thus deprecated.


merschformann (approved):
:wave:

PR #47, opened 2024-10-01
Change publish for release

sebastian-quintero:
# Description

In all of Nextmv’s repos, we use the name `release` for the workflow that publishes the "thing". This is the only place where it is called `publish`, so might as well change it for `release`. `nextroute` already uses the name `release`. We should update `nextplot` as well.


merschformann (approved):
Makes sense. Put nextplot in nextmv-io/nextplot#11.

PR #48, opened 2024-10-11
ENG-5409 Push an app from Python

sebastian-quintero:
# Description

Adds the `push()` function to the `cloud.Application` class. It _should_ be an equivalent to `nextmv app push` with the CLI.

## Testing

Kind of hard to add tests for this. I manually pushed all the community apps and ran a sample of them and it seems to work ok.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -4,6 +4,7 @@ jobs:
   python-test:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
     steps:
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -2,11 +2,12 @@ name: python test
 on: [push]
 jobs:
   python-test:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform }}
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        platform: [ubuntu-latest, windows-latest, macos-latest, macos-13]
     steps:
       - name: git clone
         uses: actions/checkout@v4
--- a/nextmv/cloud/package.py
+++ b/nextmv/cloud/package.py
@@ -1,0 +1,362 @@
+"""Module with the logic for pushing an app to Nextmv Cloud."""
+
+import glob
+import os
+import platform
+import re
+import shutil
+import subprocess
+import tarfile
+import tempfile
+from typing import Dict, List, Optional, Tuple
+
+from nextmv.cloud.manifest import FILE_NAME, Manifest, ManifestBuild, ManifestType
+from nextmv.logger import log
+
+_MANDATORY_FILES_PER_TYPE = {
+    ManifestType.PYTHON: ["main.py"],
+    ManifestType.GO: ["main"],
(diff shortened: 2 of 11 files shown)


merschformann (approved):
It would be good to have some basic tests for compress, but up to you. Thanks for the work! :hugs:

nextmv/cloud/package.py
```diff
@@ -0,0 +1,370 @@
...
+    os_agnostic_cmd = binary
+    if arguments:
+        os_agnostic_cmd += " " + " ".join(arguments)
+
+    if platform.system() == "Windows":
+        bin = "cmd"
+        args = ["/c", os_agnostic_cmd]
+    else:
+        bin = "bash"
+        args = ["-c", os_agnostic_cmd]
```
merschformann, line 253: This would be awesome to test on both platforms to make sure it works always. :relaxed:

nextmv/cloud/package.py
```diff
@@ -0,0 +1,370 @@
...
+
+    mandatory_files = _MANDATORY_FILES_PER_TYPE[manifest.type]
+    found_files = {os.path.normpath(file): True for file in present_files}
+    missing_files = [file for file in mandatory_files if file not in found_files]
+
+    if missing_files:
+        raise Exception(f"missing mandatory files: {', '.join(missing_files)}")
+
+
+def __install_dependencies(
```
merschformann, line 192: I would highly recommend evaluating the possibility to invoke `pip` directly, as we are already _within_ Python here. Something like this should do:

```python
import pip


def install(package):
    pip.main(["install", package])


# Example
if __name__ == "__main__":
    install("requests")
```

However, as I haven't confirmed that this works in all situations, I won't block. It's just that relying on a correct `PATH` setup, will fail us at some point (imho).

merschformann: (it may also simplify the code)

merschformann: Then again, this may not be safe way to invoke pip. :thinking:

sebastian-quintero: From the [official docs](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program):
> While it is implemented in Python, and so is available from your Python code via import pip, you must not use pip’s internal APIs in this way

So let’s leave it the way it is 🙏🏻 .

sebastian-quintero:
Added some simple tests for packaging

nextmv/cloud/package.py
merschformann: Yeah, saw that then as well. A little unfortunate as we then need to rely on local setup being proper, but fair enough. :relaxed:

.github/workflows/python-test.yml
```diff
@@ -4,6 +4,7 @@ jobs:
   python-test:
     runs-on: ubuntu-latest
```
merschformann: Can we broaden our testing to windows and macos in one version as well? Just to make sure that the behavior is the same (particularly for Windows). :sweat_smile: 
This should just be a matter of redefining the matrix here and the `runs-on`.

merschformann:
Thanks for the extra coverage @sebastian-quintero ! :hugs:

.github/workflows/python-test.yml
sebastian-quintero: Just did it.

nextmv/cloud/package.py
sebastian-quintero: Pat did some Windows testing, we _think_ it works. Moving forward.

PR #51, opened 2024-11-23
Apps from Models

sebastian-quintero:
This PR introduces what I am calling the "Apps from Models" logic 👉🏻  we create a Nextmv Application from a Nextmv Python Model. This means that we can have a _fully native_ experience in Python of creating a decision `Model`, pushing it to Cloud, and running it, without ever having to leave the Python session (like a script, or a notebook).

This is possible by leveraging `mlflow` and their "Models from Code" logic.


nextmv/model.py
```diff
@@ -0,0 +1,237 @@
...
+        # working with the "app from model" logic, which involves working with
+        # this `Model` class.
+        try:
+            import mlflow as mlflow
+        except ImportError:
+            # If mlflow is not installed, we install it.
+            import subprocess
+            import sys
+
+            command = [sys.executable, "-m", "pip", "install", _MLFLOW_DEPENDENCY]
```
dirkschumacher: Is that idiomatic behavior to install something without user interaction?

README.md
```diff
@@ -235,6 +235,168 @@ Write the output data after a run is completed.
...
+#### Model
+
+A decision model is a program that makes decisions, i.e.: solves decision
+problems. The model takes in an input (representing the problem data), options
+to configure the program, and returns an output, which is the solution to the
+decision problem. The `nextmv.Model` class is the base class for all models. It
+holds the necessary logic to handle all decisions.
+
+When creating your own decision model, you must create a class that inherits
+from `nextmv.Model` and implement the `solve` method.
```
merschformann, line 247: > [...] you **must** create [...]

Are we _fully_ moving to this as the new paradigm? I.e., changing all other community apps, etc. as well?

README.md
```diff
@@ -235,6 +235,168 @@ Write the output data after a run is completed.
...
+
+```python
+import nextmv
+
+class YourCustomModel(nextmv.Model):
+    def solve(self, input: nextmv.Input, options: nextmv.Options) -> nextmv.Output:
+        """Implement the logic to solve the decision problem here."""
+        pass
+
```
merschformann:
```suggestion
```

Just harmonizing trailing newlines.

README.md
```diff
@@ -235,6 +235,168 @@ Write the output data after a run is completed.
...
+```python
+import time
+from importlib.metadata import version
+
+import highspy
+import nextmv
+
+class DecisionModel(nextmv.Model):
+    def solve(self, input: nextmv.Input, options: nextmv.Options) -> nextmv.Output:
```
merschformann: Doesn't `Input` class already contain `options: Optional[Options] = None`? What is the difference to the `options` provided here? (compared to `input.options`)

README.md
```diff
@@ -235,6 +235,168 @@ Write the output data after a run is completed.
...
+
+
+model_configuration = nextmv.ModelConfiguration(
+    name="highs_model",
+    requirements=[ # Acts as a requirements.txt file.
+        "highspy==1.8.1", # Works like a line in a requirements.txt file.
+        "nextmv==0.14.0"
+    ],
+    options=options,
+)
```
merschformann, line 391: This seems like we are fully moving away from `app.yaml` for Python, which would change the experience for (all) Python developers compared to other users. Is this intended? :thinking:

README.md
```diff
@@ -253,19 +415,63 @@ Additionally, you must have a valid app in Nextmv Cloud.
...
-import os
+1. Specifying `app_dir`, which is the path to an app’s root directory. This
+   acts as an external strategy, where the app is composed of files in a
+   directory and those apps are packaged and pushed to Nextmv Cloud. This is
+   language-agnostic and can work for an app written in any language.
 
-from nextmv import cloud
+    Place the following script in the root of your app directory and run it to
+    push your app to the Nextmv Cloud. This is equivalent to using the Nextmv
+    CLI and running `nextmv app push`.
```
merschformann, line 426: I see that there is another option, but I am still missing a more CI/CD friendly way of deploying. I.e., we seem to be avoiding CLI here. If this is intentional, can we offer an entry point for nextmv-py that allows running the push _without_ adding further scripts? Something like `nextmvpy push --api-key <api-key> --app-id <app-id>` maybe?

nextmv/cloud/application.py
```diff
@@ -572,6 +573,11 @@ def new_run(
             upload_id = upload_url.upload_id
             upload_id_used = True
 
+        if options is not None:
+            for key, value in options.items():
+                if not isinstance(value, str):
+                    options[key] = json.dumps(value)
```
merschformann, line 579: <3

nextmv/model.py
```diff
@@ -0,0 +1,237 @@
...
+# the model requires and that we install and bundle with the app.
+_REQUIREMENTS_FILE = "model_requirements.txt"
+
+# When working in a notebook environment, we don’t really create a `main.py`
+# file with the main entrypoint of the program. Because the logic is mostly
+# encoded inside the `Model` class, we need to create a `main.py` file that we
+# can run in Nextmv Cloud. This file is used as that entrypoint.
+_ENTRYPOINT_FILE = "__entrypoint__.py"
+
+_MLFLOW_DEPENDENCY = "mlflow==2.18.0"
```
merschformann: I think this should be `>=` instead. Maybe it works even with older versions.
My point: we will only install mlflow if we don't find it anyway, otherwise, we will just use whatever mlflow is present. Hence, no need to pin the version.

nextmv/model.py
```diff
@@ -0,0 +1,237 @@
...
+
+            return nextmv.Output(
+                options=options,
+                solution=nextroute_output.solutions[0].to_dict(),
+                statistics=nextroute_output.statistics.to_dict(),
+            )
+        ```
+    """
+
+    def solve(self, input: Input, options: Options) -> Output:
```
merschformann: I left a comment on the readme about this function's signature. :blush:

nextmv/model.py
merschformann: It seems a little odd to install dependencies silently _without knowledge or consent of the user_. Why is this necessary?

Are we just avoiding listing `mlflow` as a dependency of `nextmv` due to its size? Couldn't it be an optional dependency? I have seen this in other packages before. It doesn't seem uncommon to me.
I.e., we would simply crash here and provide some helpful info about how to install `mlflow`. :shrug: 

E.g., `pandas` prominently does this: https://pandas.pydata.org/pandas-docs/stable/getting_started/install.html#optional-dependencies

nextmv/options.py
```diff
@@ -104,24 +151,33 @@ class Options:
...
 
         for p, param in enumerate(parameters):
             if not isinstance(param, Parameter):
                 raise TypeError(f"expected a <Parameter> object, but got {type(param)} in index {p}")
 
+            # See comment below about ipykernel adding a `-f` argument. We
+            # restrict parameters from having the name 'f' or 'fff' for that
+            # reason.
+            if param.name == "f" or param.name == "fff":
+                raise ValueError("parameter names 'f', 'fff' are reserved for internal use")
```
merschformann, line 179: This unnecessarily constrains all non-notebook use-cases from using `-f` as a shorthand, right? Potentially not the biggest loss, but it feels odd to me. :sweat_smile:

nextmv/options.py
```diff
@@ -221,9 +360,7 @@ def _parameter_value(parameter: Parameter, value: Any) -> Any:
...
         value = str(value).lower()
 
-        if value in ("true", "1", "t", "y", "yes"):
+        if value in ("true", "True", "TRUE", "1", "t", "T", "y", "Y", "yes", "Yes", "YES"):
             return True
-        if value in ("false", "0", "f", "n", "no"):
-            return False
 
-        raise argparse.ArgumentTypeError(f"invalid value for bool parameter '{parameter.name}': {value}")
+        return False
```
merschformann: A pretty relaxed approach to parsing bool maybe, but I think I am fine with it. :blush:

nextmv/options.py
```diff
@@ -221,9 +360,7 @@ def _parameter_value(parameter: Parameter, value: Any) -> Any:
 
         value = str(value).lower()
 
-        if value in ("true", "1", "t", "y", "yes"):
+        if value in ("true", "True", "TRUE", "1", "t", "T", "y", "Y", "yes", "Yes", "YES"):
```
merschformann: The different cases are not necessary. `value = str(value).lower()` will take care of that. You can eliminate any check for uppercase.

```suggestion
        if value in ("true", "1", "t", "y", "yes"):
```

merschformann (commented):
I left a number of comments, let me know what you think.

nextmv/model.py
sebastian-quintero: Moving on to setting `mlflow` as an optional dependency.

README.md
sebastian-quintero: TBD, but I think it is a good idea to unify the experience. The experience is compatible with the current apps, so I don’t see a challenge. It is mostly changing the `solve` method for the declaration of the custom class.

README.md
sebastian-quintero: Thanks!

README.md
sebastian-quintero: Yes, `input` already provides `options`. The difference would be the options used to create the input vs the options that I need at this point for solving, which might be the same. In community apps there are cases where we modify the options and use them for the output, but they are different options at the moment of input construction. It might be a 🍅 / 🍅 problem at this point, but the main idea was to comply to the `solve` method signature which is present in the community apps.

README.md
sebastian-quintero: We are not, though. I created this `ModelConfiguration` object to be able to construct an `app.yaml` manifest _programatically_, this is, without "leaving Python" (in the notebook experience). However, for a normal Python app, it still holds true that an `app.yaml` manifest is needed.

README.md
sebastian-quintero: Well, my intention for clarifying here is that people can rely on `nextmv-py` and not "know" about the CLI. I don’t know that the Python SDK should become a CLI tool as well, so I will defer that decision to others. In any case, that would  be a follow up and I would address that in a different PR.

nextmv/model.py
sebastian-quintero: Cool!

nextmv/model.py
sebastian-quintero: Responded in that comment 🙌🏻

nextmv/options.py
sebastian-quintero: Yes, I agree, it is quite odd. However, this is the workaround that I was able to find.

nextmv/options.py
sebastian-quintero: Yes, working in a Notebook with `nextmv.Options` and a bool option turned out problematic, so that is why we had to default to `False`, instead of an exception. Sorry.

nextmv/options.py
sebastian-quintero: Good catch! Thank you!

README.md
merschformann: Fair point. I still think this is an odd or non-idiomatic experience for a non-notebook use-case (particularly for harmonic and CI/CD reasons). It would be nice to discuss this before we change the recommendation in general. :blush:

README.md
merschformann: I left a reply on the other related comment [here](https://github.com/nextmv-io/nextmv-py/pull/51#discussion_r1866662127).

README.md
merschformann: I am a bit lost still. Not sure who is modifying the options provided for the run and what input construction refers to. But maybe it's not that important. Just something that seems confusing or redundant to me. :blush:

merschformann (approved):
I am still a bit concerned about some of the direction this takes us, but wouldn't block anything (particularly after our internal sync). :blush:

README.md
sebastian-quintero: Good point. I might need to change it then.

sebastian-quintero: Changed signature to inly use input.

PR #55, opened 2024-12-09
Enhancements to acceptance test models and addition of result polling method

merschformann:
# Description

Introduced new models for detailed metric analysis in acceptance tests and added a method for polling acceptance test results. Furthermore, added version and instance management for an app.

## Changes

- Introduced further models in `acceptance_test.py` to cover the returned schemas.
- Added `status` and `results` fields to the `AcceptanceTest` model (`results` nests above new models).
- Implemented `new_acceptance_test_with_result` method in `application.py` for creating and polling the result of an acceptance test.
- Added `Version` management. Create a new one, get an existing one and list all versions.
- Added `Instance` management. Create a new one, get an existing one, update an existing one and list all instances.

Resolves ENG-5639, ENG-5632

### DIFF
--- a/nextmv/cloud/__init__.py
+++ b/nextmv/cloud/__init__.py
@@ -20,7 +20,6 @@ from .account import Account as Account
 from .account import Queue as Queue
 from .account import QueuedRun as QueuedRun
 from .application import Application as Application
-from .application import Configuration as Configuration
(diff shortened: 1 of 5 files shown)


sebastian-quintero (approved):
You da best ™️

PR #59, opened 2025-01-20
Drop 3.8, Add 3.13, Use removesuffix for multicsv

sebastian-quintero:
# Description

There was a bug when stripping the `.csv` extension for `CSV_ARCHIVE` types. To use the recommended `removesuffix` function from Python, we are dropping support for version 3.8, as it was introduced in 3.9. Additionally, we are incorporating support for 3.13. These changes fall in line with the [Status of Python versions](https://devguide.python.org/versions/).

Lastly, linting was fixed for the new minimum target: 3.9

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
     steps:
       - name: git clone
         uses: actions/checkout@v4
--- a/nextmv/input.py
+++ b/nextmv/input.py
@@ -273,7 +273,7 @@ class LocalInputLoader(InputLoader):
         csv_ext = ".csv"
         for file in os.listdir(dir_path):
             if file.endswith(csv_ext):
-                stripped = file.strip(csv_ext[1:]).strip(".")  # Python 3.8 forces this, instead of using removesuffix()
+                stripped = file.removesuffix(csv_ext)
                 data[stripped] = self._load_utf8_encoded(
                     path=os.path.join(dir_path, file),
                     input_format=InputFormat.CSV,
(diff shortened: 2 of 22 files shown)


nextmv/input.py
```diff
@@ -273,7 +273,7 @@ def _load_archive(
         csv_ext = ".csv"
         for file in os.listdir(dir_path):
             if file.endswith(csv_ext):
-                stripped = file.strip(csv_ext[1:]).strip(".")  # Python 3.8 forces this, instead of using removesuffix()
+                stripped = file.removesuffix(csv_ext)
```
sebastian-quintero, line 276: Most important change is this one

.github/workflows/python-lint.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
merschformann, line 9: `3.13`

:sunglasses:

nextmv/input.py
merschformann: Something like this would also have fixed it:

```python
if file.endswith(csv_ext):
    file = file[: -len(csv_ext)]
```

However, I am good with moving forward if we don't see issues with that among our users. As this PR already shows there are some other benefits. :relaxed: 

One question: Do we need to also allow `.CSV` or even be case-insensitive? :thinking: 

Probably not. Just some thoughts. Nothing blocking.

merschformann: Expecting "case-insensitivity" among files is probably a Windows only issue. :sweat_smile:

merschformann (approved):
:tada:

nmisek (changes requested):
Looks good thank you!

nmisek (approved):

nextmv/input.py
sebastian-quintero: Didn't thought about it that much. Thank you. I can adjust if it ever comes to that.

PR #63, opened 2025-02-18
Create interface for model, and statistics instantiation

sebastian-quintero:
# Description

Creates functions for summoning a model from one of the supported third-party libraries and getting statistics from it. Here is a short example:

```python
from nextmv_sklearn.dummy import DummyRegressor, DummyRegressorOptions, DummyRegressorStatistics
from nextmv_sklearn.ensemble import (
    GradientBoostingRegressor,
    GradientBoostingRegressorOptions,
    GradientBoostingRegressorStatistics,
    RandomForestRegressor,
    RandomForestRegressorOptions,
    RandomForestRegressorStatistics,
)
from sklearn.datasets import load_diabetes

X, y = load_diabetes(return_X_y=True)

# Dummy
print("===== Dummy =====")
dum_opt = DummyRegressorOptions().to_nextmv()
dum_reg = DummyRegressor(dum_opt)
dum_fit = dum_reg.fit(X, y)
dummy_stats = DummyRegressorStatistics(dum_fit, X, y)
print("dummy prediction: ", dum_fit.predict(X[:1]))
print("dummy stats: ", dummy_stats.to_dict())

# Ensemble
print("===== Ensemble =====")
gb_opt = GradientBoostingRegressorOptions().to_nextmv()
gb_reg = GradientBoostingRegressor(gb_opt)
gb_fit = gb_reg.fit(X, y)
gb_stats = GradientBoostingRegressorStatistics(gb_fit, X, y)
print("gradient boosting prediction: ", gb_fit.predict(X[:1]))
print("gradient boosting stats: ", gb_stats.to_dict())

rf_opt = RandomForestRegressorOptions().to_nextmv()
rf_reg = RandomForestRegressor(rf_opt)
rf_fit = rf_reg.fit(X, y)
rf_stats = RandomForestRegressorStatistics(rf_fit, X, y)
print("random forest prediction: ", rf_fit.predict(X[:1]))
print("random forest stats: ", rf_stats.to_dict())
```

### DIFF
--- a/nextmv-gurobipy/nextmv_gurobipy/model.py
+++ b/nextmv-gurobipy/nextmv_gurobipy/model.py
@@ -1,0 +1,60 @@
+"""Defines gurobipy model interoperability."""
+
+import os
+
+import gurobipy as gp
+from gurobipy._paramdetails import param_details
+
+import nextmv
+
+
+def Model(options: nextmv.Options, license_path: str | None = "") -> gp.Model:
+    """
+    Creates a Gurobi model, using Nextmv options. The returned type is a
+    `gurobipy.Model` class. This means that once the Gurobi model is created,
+    it can be used as any other Gurobi model. This loader will look for the
+    `gurobi.lic` file in the provided `license_path`. If the file is not found,
+    it will not be read. This means that by default, you will be working with
+    Gurobi’s community license.
+
+    Only the parameters that are available in the Gurobi API are set. If a
+    parameter is not available, it will be skipped.
+
+    This function has some side effects that you should be aware of:
+    - It redirects the solver chatter to stderr.
+    - It sets the provider to "gurobi" in the options.
+
+    Parameters:
+    ----------
+    options: nextmv.Options
+        The options for the Gurobi model.
+
+    Returns:
+    ----------
+    gp.Model
+        The Gurobi model.
+    """
+
+    # Solver chatter is logged to stderr.
+    nextmv.redirect_stdout()
+
+    env = gp.Env(empty=True)
+
+    file_path = os.path.join(license_path, "gurobi.lic")
+    if os.path.isfile(file_path):
+        env.readParams(file_path)
+
+    env.start()
+    model = gp.Model(env=env)
+
+    gp_names = [val["name"] for val in param_details.values()]
+    for parameter in options.parameters:
+        name = parameter.name
+        if name not in gp_names:
+            continue
+
+        model.setParam(name, getattr(options, name))
+
+    options.provider = "gurobi"
+
+    return model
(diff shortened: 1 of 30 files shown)


nextmv-gurobipy/nextmv_gurobipy/model.py
```diff
@@ -0,0 +1,61 @@
...
+        The options for the Gurobi model.
+
+    Returns:
+    ----------
+    gp.Model
+        The Gurobi model.
+    """
+
+    # Solver chatter is logged to stderr.
+    nextmv.redirect_stdout()
```
merschformann, line 40: Still a bit torn about this one happening under the covers. Thanks for pointing it out in the comment though! :hugs:

nextmv-gurobipy/nextmv_gurobipy/options.py
```diff
@@ -6,6 +6,21 @@
 
 import nextmv
 
+SKIP_PARAMETERS = [
```
merschformann, line 9: Maybe I am missing something, but why are we skipping a number of these? The `InputFile` and maybe the `Tune*` ones I understand (mostly), but why the others? :thinking:

nextmv-gurobipy/nextmv_gurobipy/solution.py
```diff
@@ -0,0 +1,31 @@
+"""Defines gurobipy solution interoperability."""
+
+from typing import Optional
+
+import gurobipy as gp
+
+
+def Solution(model: gp.Model) -> Optional[dict[str, any]]:
```
merschformann: I like the simplicity of this one. :relaxed:

merschformann (approved):
This stuff is chonky. :sweat_smile: 

chonky-pr

Great and clean work! Thanks! :hugs:

nextmv-gurobipy/nextmv_gurobipy/options.py
sebastian-quintero: The CLI only ones can only be used with the Gurobi CLI directly, so even though the convenience function returns the parameter, we can’t use it from this SDK. For the `Cutoff`, the issue is that it requires a value of `+Inf` or `-Inf`, which:
1. We know is not reliable across computers.
2. Should be set according to the sense of the model. At the point of Option creation, we have not created the model yet, so we can’t tell.

IN any case, the option can always be set after the fact.

nextmv-gurobipy/nextmv_gurobipy/model.py
sebastian-quintero: The idea is that someone uses this if they are using Nextmv, so might as well guarantee chatter is logged to stderr.

nextmv-gurobipy/nextmv_gurobipy/solution.py
sebastian-quintero: Gotta create more solutions for the other libraries 😅

sebastian-quintero: Also this was a goo comment, because it made me realize that it should be called `ModelSolution`.

PR #68, opened 2025-03-03
Adds support for creating and deleting apps

merschformann:
# Description

Adds support for creating new applications and deleting existing ones.

## Changes

- Adds `Application.new(...)` for creating a new application
- Adds `app.delete()` for deleting an existing _Application_.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -871,6 +871,56 @@ class Application:
 
         return Instance.from_dict(response.json())
 
+    @staticmethod
+    def new(
+        client: Client,
+        name: str,
+        id: Optional[str] = None,
+        description: Optional[str] = None,
+    ) -> "Application":
+        """
+        Create a new application.
+
+        Args:
+            client: Client to use for interacting with the Nextmv Cloud API.
+            name: Name of the application.
+            id: ID of the application. Will be generated if not provided.
+            description: Description of the application.
+
+        Returns:
+            The new application.
+        """
+
+        payload = {
+            "name": name,
+        }
+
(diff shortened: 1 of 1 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -871,6 +871,56 @@ def new_instance(
 
         return Instance.from_dict(response.json())
 
+    @staticmethod
```
sebastian-quintero, line 874: If it is a static method, shouldn’t we maybe have it as a convenience function outside of the class? Have a func like `def new_application(...) -> Application`? That way we don’t need the `"Application"` between quotes? After all, we are not using anything from the class, so maybe the method doesn’t belong there.

sebastian-quintero (approved):
Left a comment for you to consider.

nextmv/nextmv/cloud/application.py
merschformann: Yeah, it really depends on where we want it to live. As a static on Application works for me, but an independently defined func is fine for me too. That would go on `cloud` package then, right?

So, either

```python
app = Application.new(...)
```

or 

```python
app = cloud.new_application(...)
```

Regarding the quotes around the type hint, that is still quite idiomatic Python, I think (based on [PEP 484](https://peps.python.org/pep-0484/#forward-references)). That doesn't mean that I lean towards it though. I don't have a favorite. :blush: 

Which one do you lean towards?

sebastian-quintero: Either works for me 🤷🏻‍♂️, that is why I approved. Whatever you decide, as you are the author!

PR #71, opened 2025-03-27
Adds app exists check

merschformann:
# Description

Adds more convenient way of checking whether an app exists.

## Changes

* Adds `nextmv.cloud.Application.exists(client, app_id)` for checking whether an app exists.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -908,6 +908,33 @@ class Application:
 
         return Application(client=client, id=response.json()["id"])
 
+    @staticmethod
+    def exists(
+        client: Client,
+        id: str,
+    ) -> bool:
+        """
+        Check if an application exists.
+
+        Args:
+            client: Client to use for interacting with the Nextmv Cloud API.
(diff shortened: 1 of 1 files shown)


nmisek (approved):

nextmv/nextmv/cloud/application.py
```diff
@@ -908,6 +908,46 @@ def new(
 
         return Application(client=client, id=response.json()["id"])
 
+    @staticmethod
+    def exists(
```
sebastian-quintero, line 912: I wonder is we should use this func in places like post init? To instantiate the class only if the app exists

sebastian-quintero (approved):

nextmv/nextmv/cloud/application.py
merschformann: I wondered about it too. Then again, would we then also do the same when instantiating an instance, version, etc.? :thinking: I am not sure.
Also, I was wondering whether to also add exists checks for instance, version, etc. Not sure where it would fit best and whether necessary though.

PR #74, opened 2025-04-01
Improve the polling strategy

sebastian-quintero:
# Description

Improves the polling strategy using exponential backoff with jitter, as opposed to a more constant polling. Makes the default values more generous to poll for more time on runs that take their sweet time to complete (like when a customer first creates an account).

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -1,6 +1,7 @@
 """This module contains the application class."""
 
 import json
+import random
 import shutil
 import time
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -37,24 +38,57 @@ class DownloadURL(BaseModel):
...
+    * i is the retry (poll) number.
+    * Uniform is the uniform distribution.
+
+    Note that the sleep duration is capped by the `max_delay` parameter.
+    """
+
+    backoff: float = 0.9
+    """
+    Exponential backoff factor, in seconds, to use between polls. Leave this at
+    1 to poll at a constant rate.
```
merschformann:
```suggestion
    Exponential backoff factor, in seconds, to use between polls.
```

I think the comment was outdated and maybe not necessary? Just a recommendation though. Feel free to add more context. :blush:

nextmv/nextmv/cloud/application.py
```diff
@@ -37,24 +38,57 @@ class DownloadURL(BaseModel):
...
     """Maximum duration of the polling strategy, in seconds."""
-    max_tries: int = 20
+    max_tries: int = 100
     """Maximum number of tries to use."""
+    jitter: float = 1
+    """
+    Jitter to use for the polling strategy. A uniform distribution is sampled
+    between 0 and this number. The resulting random number is added to the
+    delay for each poll, adding a random noise. Leave this at 0 to avoid using
+    random jitter.
```
merschformann:
```suggestion
    delay for each poll, adding a random noise. Set this to 0 to avoid using
    random jitter.
```

Just while I was at it, since it is technically not at 0 by default.

merschformann (approved):
Thanks and very clean! :hugs: :broom: 
Just left two very tiny suggestions. Feel free to ignore though.

sebastian-quintero:
@merschformann Thank you! I committed your suggestions.

PR #76, opened 2025-04-16
Add functionality to track a run (use external runs)

sebastian-quintero:
# Description

Adds two new methods to the `cloud.Application`:
* `track_run`: tracks a new run (convenience function for registering an external run).
* `track_run_with_result`: calls track a new run but executed polling (which should be almost immediate).

For this, a new `TrackedRun` class is introduced. The original API design for `external_runs` is still there, and this is just a convenient method that combines uploading a large input and obtaining the `upload_id` to be able to track the external run.


merschformann (approved):
:tada:

PR #79, opened 2025-04-18
Support running with nextmv.Input and nextmv.Options

sebastian-quintero:
# Description

The `Application.new_run` and `Application.new_run_with_result` (by extension) were updated to support `nextmv.Input` as an input in the method. When this object is used, the data is extracted from `Input.data`. Right now, the only supported `input_format` is `JSON`, given that we don’t support stuff like `csv-archive` in the SDK yet.

More updates were done to support `nextmv.Options` as run options as well.

Lastly, `configuration` and `external_result` were made more flexible to accept a `dict` as well as the classes.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -805,11 +805,11 @@ class Application:
         ----------
         input: Union[Input, dict[str, Any], BaseModel, str]
             Input to use for the run. This can be a `nextmv.Input` object,
-            dict, BaseModel or string. If `nextmv.Input` is used, then the
+            `dict`, `BaseModel` or `str`. If `nextmv.Input` is used, then the
             input is extracted from the `.data` property. Note that for now,
-            the only supported `.input_format` for this method is `JSON`. If a
-            string is provided, the input will be uploaded, associating it with
-            an `upload_id`. In case a dict is used, it will be used as is.
+            `InputFormat.CSV_ARCHIVE` is not supported as an
+            `input.input_format`. If an input is too large, it will be uploaded
+            with the `upload_large_input` method.
         instance_id: Optional[str]
             ID of the instance to use for the run. If not provided, the default
             instance ID associated to the Class (`default_instance_id`) is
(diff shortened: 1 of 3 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -784,51 +785,93 @@ def new_instance(
...
-                with.
-            external_result: External result to use for the run, if this is an
-                external run.
+        Parameters
+        ----------
+        input: Union[Input, dict[str, Any], BaseModel, str]
+            Input to use for the run. This can be a `nextmv.Input` object,
+            dict, BaseModel or string. If `nextmv.Input` is used, then the
+            input is extracted from the `.data` property. Note that for now,
+            the only supported `.input_format` for this method is `JSON`. If a
```
merschformann: Theoretically, we support UTF-8 input (not only json). Not sure whether we want to relax here to support that already?

nextmv/nextmv/cloud/application.py
```diff
@@ -784,51 +785,93 @@ def new_instance(
...
+        ----------
             requests.HTTPError: If the response status code is not 2xx.
+            ValueError:
+                If the `input` is of type `nextmv.Input` and the
+                `.input_format` is not `JSON`.
+                If the final `options` are not of type `dict[str,str]`.
         """
 
-        input_size = 0
+        input_data = None
```
merschformann, line 862: Won't input data just be `None` in case of a pure string input? :thinking: 
I am not sure how the `str` input type is handled here (as mentioned in comment above). Maybe I am just missing something. Pure string should probably just passed on, since it is acceptable by the platform.

merschformann (approved):
Nice work! Thanks! :hugs: 
I had a question about certain case in a func, but don't want to be blocking in case I am just missing something. :relaxed:

nextmv/nextmv/cloud/application.py
merschformann: Oh, I see now. The plain str is intended to be an upload ID. Didn't get that immediately from the comment. Hmmm, that is a bit sad that it will conflict a bit with having a CSV input for example or any other pure text input. Can we change that around and pass the upload ID as `upload_id=<id>` instead maybe (where `input` would be ignored if upload id provided)? Or is it too late for that?

nextmv/nextmv/cloud/application.py
sebastian-quintero: Good point. I will change the logic to rather say "csv-archive is not supported for now". Thanks!

nextmv/nextmv/cloud/application.py
sebastian-quintero: Yeah I think I messed something up. Str input should just be passed on and depending on size be uploaded, same as input. I'll take a second look tonight.

sebastian-quintero: Ok, I adjusted it in the new revision. `dict` and `str` behave the same way now, as they should.

nextmv/nextmv/cloud/application.py
sebastian-quintero: Adjusted in the new revision.

PR #80, opened 2025-04-21
Adds stop condition to polling mechanism

sebastian-quintero:
# Description

Extends the polling mechanism to allow for a stop condition to be passed in via a callback. This enables the polling process to be stopped based on external factors or conditions, providing more control over long-running operations. Also fixes a type annotation issue for the polling function.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -4,6 +4,7 @@ import json
 import random
 import shutil
 import time
+from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import datetime
 from typing import Any, Optional, Union
@@ -40,7 +41,8 @@ class DownloadURL(BaseModel):
     """URL to use for downloading the file."""
 
 
-class PollingOptions(BaseModel):
+@dataclass
+class PollingOptions:
     """
     Options to use when polling for a run result.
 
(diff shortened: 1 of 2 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -4,6 +4,7 @@
 import random
 import shutil
 import time
+from collections.abc import Callable
```
merschformann, line 7: Hmm, I always use this:

```python
from typing import Callable
```

Which one is preferable? :thinking:

merschformann:
```suggestion
from typing import Callable
```

:point_up:  Just in case you want to quickly replace it. But again, ignore if this is the more fitting type. :relaxed:

merschformann (approved):
Thanks for the work! Looks good. I only had a simple type hinting question. :relaxed:

nextmv/nextmv/cloud/application.py
sebastian-quintero: Yeah so according to the [Python docs](https://docs.python.org/3/library/typing.html#annotating-callable-objects):
> Functions – or other [callable](https://docs.python.org/3/glossary.html#term-callable) objects – can be annotated using [collections.abc.Callable](https://docs.python.org/3/library/collections.abc.html#collections.abc.Callable) or deprecated [typing.Callable](https://docs.python.org/3/library/typing.html#typing.Callable). Callable[[int], str] signifies a function that takes a single parameter of type [int](https://docs.python.org/3/library/functions.html#int) and returns a [str](https://docs.python.org/3/library/stdtypes.html#str).

This means that `typing.Callable` is deprecated. Moreover, inspecting the source code, `typing.Callable` has a deprecated annotation and simply points to `collections.abc.Callable`.

merschformann: Interesting, good to know. :+1:

PR #82, opened 2025-04-25
Add `nextmv.Option` and deprecate `nextmv.Parameter`

sebastian-quintero:
# Description

Replace `nextmv.Parameter` 👉🏻 `nextmv.Option`. `nextmv.Parameter` is not deleted, just marked as deprecated. All methods in that class are also marked as deprecated.

All the methods in `Options` that used `Parameters` are deprecated, and new methods supporting `Option` were introduced:
- `from_options_dict`
- `options_dict`

### DIFF
--- a/nextmv/tests/test_options.py
+++ b/nextmv/tests/test_options.py
@@ -24,7 +24,7 @@ class TestOptions(unittest.TestCase):
     assume that the script is one level up.
     """
 
-    test_scripts = [1, 2, 3, 4, 5, 6, 7, "_deprecated"]
+    test_scripts = ["1", "2", "3", "4", "5", "6", "7", "_deprecated"]
     """These are auxiliary scripts that are used to test different scenarios of
     instantiating an `Options` object."""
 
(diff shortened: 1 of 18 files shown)


nextmv/tests/test_options.py
```diff
@@ -24,7 +24,7 @@ class TestOptions(unittest.TestCase):
     assume that the script is one level up.
     """
 
-    test_scripts = [1, 2, 3, 4, 5, 6, 7]
+    test_scripts = [1, 2, 3, 4, 5, 6, 7, "_deprecated"]
```
merschformann:
```suggestion
    test_scripts = ["1", "2", "3", "4", "5", "6", "7", "_deprecated"]
```

Can we convert it to just strings? I know it works, is only in a test and doesn't really matter, but the mixed types in the list kind of send shivers down my spine. :joy: 

But yeah, feel free to ignore. Mainly noticing tbf. :blush:

merschformann (approved):
Nice work! Thanks for taking care! :hugs:

nextmv/tests/test_options.py
sebastian-quintero: Yep, sorry about that.

PR #83, opened 2025-04-25
Introduce write, load and deprecate write_local, load_local

sebastian-quintero:
# Description

Introduces the following functions:

- `nextmv.write` ➡️ replaces `nextmv.write_local` (which has been marked as deprecated).
- `nextmv.load` ➡️  replaces `nextmv.loadl_local` (which has been marked as deprecated).


merschformann (approved):
I remember a discussion about this. :thinking: 
Anyway, great work! Thank you! :hugs:

PR #84, opened 2025-04-26
Add console_url to run information and result

sebastian-quintero:
# Description

The `console_url` can not be retrieved from either the `RunResult` or the `RunInformation`, with the following methods:

- `new_run_with_result`
- `run_result`
- `run_result_with_polling`
- `run_metadata`
- `track_run_with_result`

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -1326,7 +1326,10 @@ class Application:
             endpoint=f"{self.endpoint}/runs/{run_id}/metadata",
         )
 
-        return RunInformation.from_dict(response.json())
+        info = RunInformation.from_dict(response.json())
+        info.console_url = self.__console_url(info.id)
+
+        return info
 
     def run_logs(self, run_id: str) -> RunLog:
         """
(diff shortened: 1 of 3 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -1766,6 +1771,11 @@ def __update_app_binary(
                 )
             )
 
+    def __console_url(self, run_id: str) -> str:
+        """Auxiliary method to get the console URL for a run."""
+
+        return f"https://cloud.nextmv.io/app/{self.id}/run/{run_id}?view=details"
```
merschformann: Doesn't this (kind of) demand support for other endpoints/environments as well? Particularly, when considering current roadmap and priorities (+testing).

Maybe the `nextmv.cloud.Client` could have a custom `console_url` (in addition to the API `url`) that we use here?

merschformann (approved):
Nice!

I had a medium comment that I don't want to mark blocking, as it can also be a follow-up. However, it seems important to have eventually.

nextmv/nextmv/cloud/application.py
sebastian-quintero: This is why you are my always trusted reviewer, because you catch things that I overlook for doing things too fast and loose 😄 .

merschformann: Naaww. :heart:

PR #85, opened 2025-04-28
Adds options to the Manifest

sebastian-quintero:
# Description

Adds a new property to the `Manifest`: `options`. This is a list of a new object, called `ManifestOption`. Convenience methods are introduced to `ManifestOption` and `Manifest` to be able to extract `nextmv.Options` from them, and viceversa: instantiating if one already has the options.

For the "Apps from Models" experience (notebooks), the `__entrypoint__.py` is modified so that the options are consumed from the manifest options as opposed to the underlying model configuration options.


merschformann (approved):
Thank you! :partying_face:

PR #86, opened 2025-04-28
Make adjustments to tracked run

sebastian-quintero:
# Description

- For tracking the input, we were using `nextmv.Input.to_dict`, which resulted in the input having `options`, `data`, and `input_format`. Instead, we just use `nextmv.Input.data` as an input, to have the actual data as input.
- For tracking the output, the `dict` representation of the `nextmv.Output` is adjusted to stop showing `null` and show empty objects or lists instead, where applicable. This is in line with how `nextmv.write` does it.


merschformann (approved):
Makes sense! :+1:

PR #87, opened 2025-05-06
Add scenario tests to the application

sebastian-quintero:
# Description

Adds new functionality for the `cloud.Application`:

- `new_scenario_test`: create a new scenario test.
- `scenario_test`: get a scenario test.
- `list_scenario_tests`: list scenario tests.
- `delete_scenario_tests`: deletes a scenario test.


nextmv/nextmv/cloud/application.py
```diff
@@ -230,18 +241,40 @@ def delete_batch_experiment(self, batch_id: str) -> None:
...
+        scenario_test_id: str
+            ID of the scenario test.
+
+        Raises
+        ------
+        requests.HTTPError
+            If the response status code is not 2xx.
+        """
+
+        _ = self.delete_batch_experiment(batch_id=scenario_test_id)
```
merschformann: Why is it btw that we are explicitly throwing away the return value of some functions (`_ = ...`). This one does not even have a return value, right? (it's a `-> None` I think)
Is it a style convention? :thinking:

nextmv/nextmv/cloud/application.py
```diff
@@ -409,6 +447,53 @@ def list_instances(self) -> list[Instance]:
...
+
+        Raises
+        ------
+        requests.HTTPError
+            If the response status code is not 2xx.
+        """
+
+        response = self.client.request(
+            method="GET",
+            endpoint=f"{self.experiments_endpoint}/batch?",
```
merschformann: Is the trailing `?` necessary here? The underlying `requests` should automatically add it when adding params, right? :thinking: 
Not important though if it works. It just looks a bit odd to me.

nextmv/nextmv/cloud/application.py
```diff
@@ -1048,6 +1297,124 @@ def new_run_with_result(
...
+
+        if len(scenarios) < 1:
+            raise ValueError("At least one scenario must be provided")
+
+        # Save all the information needed by scenario.
+        input_sets = {}
+        scenarios_by_id = {}
+        instances = {}
+        for scenario_ix, scenario in enumerate(scenarios):
+            scenario_id = f"scenario-{scenario_ix}" if scenario.scenario_id is None else scenario.scenario_id
```
merschformann: Do we need to verify that we are generating a unique ID here? I.e., `scenario.scenario_id` of one scenario could technically already be defined as "scenario-0" which we would generate for the first if it wouldn't have an ID. :thinking: 

Probably a bit overcautious for now. I only use scenario testing occasionally and don't have a good sense for the use case of this `nextmv-py` functionality yet. :blush:

nextmv/nextmv/cloud/run.py
```diff
@@ -164,7 +164,7 @@ class RunConfiguration(BaseModel):
     """Run type configuration for the run."""
     secrets_collection_id: Optional[str] = None
     """ID of the secrets collection to use for the run."""
-    queuing: Optional[RunQueueing] = None
+    queuing: Optional[RunQueuing] = None
     """Queueing configuration for the run."""
```
merschformann:
```suggestion
    """Queuing configuration for the run."""
```

Here is another one. :blush:

nextmv/nextmv/cloud/scenario.py
```diff
@@ -0,0 +1,206 @@
...
+from typing import Any, Optional, Union
+
+
+@dataclass
+class ScenarioConfiguration:
+    """
+    Configuration for a scenario.
+
+    You can define multiple values for a single option, which will result in
+    multiple scenarios being created. For example, if you have a configuration
```
merschformann: > You can define multiple values for a single option, which will result in **_multiple scenarios_** being created

I am actually not fully clear on what exactly we refer to as a "scenario". It seems like console UI may have a different understanding than what this comment states. This comment seems to suggest that any fine-granular / atomic / cannot be broken down further unit is a scenario.

I actually agree with that description, but am not sure whether we apply the terminology consistently. :thinking:

merschformann (approved):
Nice work! :hugs: 

I only left some minor comments and thoughts. Nothing really needs to be done for these. :blush:

nextmv/nextmv/cloud/application.py
sebastian-quintero: No convention here, just me being stupid. I fixed it.

nextmv/nextmv/cloud/application.py
sebastian-quintero: Leftover from some stuff I was trying. Nice catch, I removed it.

nextmv/nextmv/cloud/run.py
sebastian-quintero: Thanks

nextmv/nextmv/cloud/scenario.py
sebastian-quintero: Good catch, I meant to say "runs" here, not scenario/

nextmv/nextmv/cloud/application.py
sebastian-quintero: Good point. Better be safe than sorry. I will create a function for handling scenario IDs creation and checking that they result in unique keys.

sebastian-quintero: I created the `_scenarios_by_id` that creates the IDs of the scenarios and makes sure that they are unique.

PR #92, opened 2025-05-14
Addresses issue with multiple stdout redirects

merschformann:
# Description

Fixes an issue where multiple (unintentional) `nextmv.redirect_stdout()` calls would loose the original `sys.stdout` reference that was originally set. This PR blocks successive redirects and makes sure the original `sys.stdout` is safe. The implementation does not support multi-threading (yet).

## Changes

* Fixes issue with multiple successive `nextmv.redirect_stdout()` calls.

### DIFF
--- a/nextmv/nextmv/logger.py
+++ b/nextmv/nextmv/logger.py
@@ -3,13 +3,17 @@
 import sys
 
 __original_stdout = None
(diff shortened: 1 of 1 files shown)


sebastian-quintero (approved):
Thanks!

PR #102, opened 2025-05-28
Allows requirements as code

merschformann:
# Description

Allows the user to specify requirements as code. I.e., the user can now setup their app **_without_** having to create _app.yaml_ and _requirements.txt_ files.

This allows to only have the following files / structure:

```txt
.
├── main.py           # The actual app / model code
└── push.py           # Python script to define and push the app (see below)
```

Example of a **_push.py_** file (next to the _main.py_ file containing the actual app/model code):

```python
import datetime
import os

from nextmv import cloud

# App directory is the same as this file.
app_dir = os.path.dirname(os.path.abspath(__file__))

# Create and push the app to the cloud.
client = cloud.Client(api_key=os.getenv("NEXTMV_API_KEY"))
app = cloud.Application(client=client, id=os.getenv("<app-id>"))
app.push(
    app_dir=app_dir,
    verbose=True,  # See some output from the push process.
    manifest=cloud.Manifest(
        files=[
            "main.py",  # The main file of the app.
        ],
        python=cloud.ManifestPython(
            pip_requirements=[
                "nextmv==0.28.1",
                "ortools==9.12.4544",
            ],
        ),
        type=cloud.ManifestType.PYTHON,
        runtime=cloud.ManifestRuntime.PYTHON,
    ),
)

# Create a new version and update the staging instance to use it.
version_id = f"sandbox-auto-{datetime.datetime.now().strftime('%Y-%m-%dT%H-%M-%S')}"
app.new_version(id=version_id, name=f"Auto version {version_id}")
app.update_instance(id="staging", name="staging", version_id=version_id)
```

The example above does the following:

* Make sure to push from the directory of the app independent of where it was called from.
* Use a `NEXTMV_API_KEY` from env for auth.
* Push the code for `<app-id>` app.
* Bundle `main.py` (which contains the model code).
* Define all requirements.
* Set the app type to Python.
* Set the runtime to one suiting Python apps.
* Automatically create a version for the pushed code.
* Update the `staging` instance to use the new version.

## Changes

* Allows the user to fully define and push the app in Python.

### DIFF
--- a/docs/nextmv/tutorials/applications/push.md
+++ b/docs/nextmv/tutorials/applications/push.md
@@ -10,23 +10,28 @@ following minimum requirements:
 
 1. Must read from either `stdin` or a file.
 2. Must write results either to `stdout` or a file.
-3. Must contain an [`app.yaml` manifest][app-manifest] in the root of your app
-project.
+3. Depending on the push strategy (see below), may require an `app.yaml`
+   manifest file and an entrypoint (`main.py`, `main`, `main.jar`, etc.).
 
 To push an app to Nextmv Cloud using the `nextmv` Python SDK, you can use one
 of the following strategies:
 
-* [External strategy][external-strategy]: push an app that is composed of files
-      in a directory.
+* [Directory strategy][directory-strategy]: push an _existing_ app that is
+      composed of files in a directory.
+* [Python strategy][python-strategy]: describe and push an app purely in
+      Python (without the need for an [`app.yaml` manifest][app-manifest]
+      for example).
 * [Internal strategy][internal-strategy]: push an app that is created from a
   [`nextmv.Model`][model].
 
-## External strategy
+## Directory strategy
 
-The external strategy consists of specifying `app_dir`, which is the path to an
+The directory strategy consists of specifying `app_dir`, which is the path to an
 app’s root directory. The app is composed of files in a directory and those
-apps are packaged and pushed to Nextmv Cloud. This is language-agnostic and can
-work for an app written in any language.
+apps are packaged and pushed to Nextmv Cloud. This is language-agnostic and
+works for an app written in any language. It is particularly useful
+when the app is already structured as a directory with the necessary files (like
+apps cloned from [community apps][community-apps]).
 
 Place the following script in the root of your app directory and run it to push
 your app to the Nextmv Cloud. This is equivalent to using the [Nextmv CLI][cli]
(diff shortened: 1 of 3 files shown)


nmisek (approved):

nextmv/nextmv/cloud/manifest.py
```diff
@@ -299,7 +301,7 @@ class ManifestPython(BaseModel):
     'requirements.txt'
     """
 
-    pip_requirements: Optional[str] = Field(
+    pip_requirements: Optional[Union[str, list]] = Field(
```
sebastian-quintero: Can you specify list[str] to be more precise?

sebastian-quintero (approved):
Three nits:
1. The comment I posted
2. Are we checking anywhere that main.py exists and throwing an exception in case it doesn't?
3. Can you add a little something to the docs (tutorials), that reflects this change? Maybe in the pushing an app section.

nextmv/nextmv/cloud/manifest.py
merschformann: Yeah, I thought about it but for some reason didn't add it. :confused: 
I added it. Thanks for the reminder! :heart:

merschformann:
Quick response @sebastian-quintero : 

1. The comment I posted
    * Done. Thanks for the reminder! :heart: 
2. Are we checking anywhere that main.py exists and throwing an exception in case it doesn't?
   * I am not sure, but the code that handles the directory push is the same as the one handling this "Python push". So, I am not sure whether it should be in scope for this PR. I can add it if missing though.
3. Can you add a little something to the docs (tutorials), that reflects this change? Maybe in the pushing an app section.
    * Yeah, good call. I changed the relevant section around a bit. Let me know what you think. :blush:

sebastian-quintero (approved):
I would combine dir strategy and Python strategy in one, maybe, to avoid offering too many options, which might lead to confusion. In there, would then clarify that you can have requirements and app.yaml in the dir or use Python constructs to specify them. I like dir strategy more than external. And maybe we rename internal to "Model" strategy? The main clarification is we create an app on the fly from a Model.

merschformann:
I gave it another restructure @sebastian-quintero . I combined the "dir" approach and "python" approach (was a terrible name) into one "file strategy" that has a directory based and a code based option. I also renamed the external strategy to object strategy. I also like your model name, but I think "object" is spot on about what is happening. Maybe too bold though. :shrug: 

I hope this makes sense. :blush:

docs/nextmv/tutorials/applications/push.md
```diff
@@ -10,23 +10,34 @@ following minimum requirements:
 
 1. Must read from either `stdin` or a file.
 2. Must write results either to `stdout` or a file.
-3. Must contain an [`app.yaml` manifest][app-manifest] in the root of your app
-project.
+3. Depending on the push strategy (see below), may require an `app.yaml`
```
sebastian-quintero, line 13: Sorry for dragging this further, but really the app.yaml is non-negotiable, so it is not correct that an app "may" need a manifest. It always does. But it is ok.

docs/nextmv/tutorials/applications/push.md
```diff
@@ -10,23 +10,34 @@ following minimum requirements:
...
+## File strategy
 
-The external strategy consists of specifying `app_dir`, which is the path to an
-app’s root directory. The app is composed of files in a directory and those
-apps are packaged and pushed to Nextmv Cloud. This is language-agnostic and can
-work for an app written in any language.
+The file strategy assumes that the model exists as source code on disk. We can
+distinguish between two types of file strategies:
+
+* **Directory based**: the app is described with a [`app.yaml`][app-manifest]
```
sebastian-quintero: An instead of a?

sebastian-quintero (approved):
Lgtm

docs/nextmv/tutorials/applications/push.md
merschformann: It is. That is the point of this PR. It allows the user to exclusively define everything in a say `push.py` file. I am doing this to make the file-based strategy more convenient for users of the model / object strategy. We are running into a number of issues still with pickling and other things. So, this is an attempt to offer a more stable but almost similar option (except for still having to have the `main.py` - push could still be defined within that too, but that feels unclean). :blush: 

I'll point that out in the PR description as well. And I might also do a tree in the docs to make it clearer. :blush:

sebastian-quintero:
Merge merge merge

PR #105, opened 2025-06-04
Adds multi-way option merge

merschformann:
# Description

Adds functionality for merging multiple option sets at once. Furthermore, allows the user to skip parsing if further options will be added later after the merge was done.

## Changes

* Adds `Options.merge_all(...)` function for merging multiple options sets at once.
  * E.g.: `merged_options = Options.merge_all(option_set1, option_set2, option_set3)`
* Adds `skip_parse` parameter to option merge functions to allow lazy option merging and delayed (explicit) parsing.

### DIFF
--- a/nextmv/nextmv/options.py
+++ b/nextmv/nextmv/options.py
@@ -676,15 +676,15 @@ class Options:
         if not options:
             raise ValueError("at least one Options instance is required to merge")
 
-        merged = []
+        merged_options_list = []
         for opt in options:
             if opt.PARSED:
                 raise RuntimeError(
                     "options have already been parsed, cannot merge. See `Options.parse()` for more information."
                 )
-            merged += opt.options
+            merged_options_list += opt.options
 
-        merged_options = cls(*merged)
+        merged_options = cls(*merged_options_list)
 
         if not skip_parse:
             merged_options._parse()
--- a/nextmv/nextmv/options.py
+++ b/nextmv/nextmv/options.py
@@ -587,7 +587,8 @@ class Options:
         Parameters
         ----------
         new : Options
-            The new options to merge with the current options.
+            The new options to merge with the current options. At least one new option set
+            is required to merge. Multiple `Options` instances can be passed.
         skip_parse : bool, optional
             If True, the merged options will not be parsed after merging. This is useful
             if you want to merge further options after this merge. The default is False.
--- a/nextmv/nextmv/options.py
+++ b/nextmv/nextmv/options.py
@@ -627,7 +627,8 @@ class Options:
 
         self.options += new.options
 
-        self._parse()
+        if not skip_parse:
+            self._parse()
 
         return self
 
--- a/nextmv/nextmv/options.py
+++ b/nextmv/nextmv/options.py
@@ -575,7 +575,7 @@ class Options:
 
         self._parse()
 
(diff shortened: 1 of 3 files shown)


nextmv/nextmv/options.py
```diff
@@ -624,10 +627,70 @@ def merge(self, new: "Options") -> "Options":
...
         self.options += new.options
 
-        self._parse()
+        if not skip_parse:
+            self._parse()
 
         return self
 
+    @classmethod
+    def merge_all(cls, *options: "Options", skip_parse: bool = False) -> "Options":
```
sebastian-quintero: I think that we should change this to go with one of the following:
1. Make this a proper method on the class and merge all incoming options to the current class. Like "merge", but receiving more than one options object. 
2. Make it a standalone function outside of the class. 

The reason is that it being a classmethod I think is unnecessary. Id it were up to me, I would go with 1 for consistency and maybe change the name to "merge_multiple".

nextmv/nextmv/options.py
```diff
@@ -624,10 +627,70 @@ def merge(self, new: "Options") -> "Options":
...
+        """
+
+        if not options:
+            raise ValueError("at least one Options instance is required to merge")
+
+        merged_options_list = []
+        for opt in options:
+            if opt.PARSED:
+                raise RuntimeError(
+                    "options have already been parsed, cannot merge. See `Options.parse()` for more information."
```
sebastian-quintero: I would add at least the index of the options object that is parsed to the msg for better debugging.

nextmv/nextmv/options.py
```diff
@@ -624,10 +627,70 @@ def merge(self, new: "Options") -> "Options":
...
+                )
+            merged_options_list += opt.options
+
+        merged_options = cls(*merged_options_list)
+
+        if not skip_parse:
+            merged_options._parse()
```
sebastian-quintero: Can't we call "parse()"??

sebastian-quintero (approved):
Can we also add docs for this in the options tutorial? There is a part on merging options, we can maybe add a new section called merging multiple options together or elaborate more on the existing section.

nextmv/nextmv/options.py
merschformann: Good call. :+1:

nextmv/nextmv/options.py
merschformann: Adding it to the class / object feels a bit weird, as there is already a way to merge options. Let me check whether there is a non-breaking way to extend it though.

Regarding the static method vs. standalone: I don't have a strong opinion. However, we are already offering other methods that are similarly tied to the class definition as static methods on the class, so, I actually feel that this is more in line with the style that was used so far.

I'll try to remove the method though and merge it with the other one.

merschformann:
> Can we also add docs for this in the options tutorial? There is a part on merging options, we can maybe add a new section called merging multiple options together or elaborate more on the existing section.

You are talking about [this example](https://nextmv-py.readthedocs.io/en/latest/nextmv/tutorials/options/), correct? I am not sure whether adding 3 option sets wouldn't overcomplicate the example. :thinking: However, happy to do it, if it seems helpful. :blush:

sebastian-quintero (approved):
Nice alternative! Yeah, I think all of this is the cleanest way.

PR #106, opened 2025-06-10
Reduces size of serialized JSON by default

merschformann:
# Description

This PR aligns the JSON serialization in the project while also aiming to reduce the size of serialized JSON as much as possible (by removing all unnecessary whitespace).

## Changes

- Introduces `serialization.serialize_json(...)` for common / aligned serialization.
- Changes default serialization to not use indentation and minimal whitespace.
  - This improves working with large inputs, assets, etc.

Resolves ENG-6118

### DIFF
--- a/docs/nextmv/tutorials/output.md
+++ b/docs/nextmv/tutorials/output.md
@@ -40,6 +40,22 @@ The `solution` property of the output is a dictionary that represents the
 output data. The `statistics` property can be a `nextmv.Statistics` object, or
 a dictionary.
 
+By default, `nextmv.Output` serializes `JSON` using pretty printing. If you want
+to change the serialization behavior, you can pass the `json_configurations`
+parameter. The provided values are passed to the underlying `json.dumps`
+method. For example, to get compressed output, you can set:
+
+```python
+output = nextmv.Output(
+    # ...
+    json_configurations={
+        "indent": None,  # No indentation for compact output
+        "separators": (",", ":")  # Use compact separators
+    },
+    # ...
+)
+```
+
 ## `CSV` output
 
 Work with one, or multiple, `CSV` files. In the `solution` property of the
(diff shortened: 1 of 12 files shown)


nextmv/nextmv/serialization.py
```diff
@@ -0,0 +1,67 @@
+import datetime
+import json
+from typing import Any, Union
+
+
+def serialize_json(
+    obj: Union[dict, list],
+    json_configurations: dict[str, Any] = None,
+) -> str:
```
sebastian-quintero: Doesn’t this fit in a single line?

sebastian-quintero (approved):
Can you please add the `serialization.py` to the docs technical reference? There are 2 things you need to do:
- Add the actual page under `docs/nextmv/reference` (new page is `serialization.md`). You can use one of the other pages as reference.
- Add the page to the index in the `mkdocs.yml` file. I would add it after `output.py` in the `Reference` section, as it si purely technical reference.

It would also be awesome if you can add a new tutorial on "Serialization", where you briefly explain how to use default vs. custom serialization. Ideally, we can add short docs as soon as we create the feature so we can keep up with them.

I think we are changing the serialization to indent bu default as well.

nextmv/nextmv/serialization.py
merschformann: That was just for readability (my intuition). But happy to change it back.

merschformann:
@sebastian-quintero : I am not sure whether we want to export the serialization functionality. `serialization.py` was meant to be used internally, to make sure we have a common default (or maybe multiple) for serialization.

Do we want to expose the serialization? :thinking: 

Currently, serialization can be controlled via the `json_configurations`, which are meant to be in line with the `csv_configurations` parameter. I will give this another round after our internal discussion anyway though. :blush:

sebastian-quintero:
It is already a "public" function in that module, as it is `serialize_json`, not `_serialize_json`, that is why I suggested it. I would either:
- Add the underline so we don’t document it.
- Add the reference as suggested.

Reference discussion apart, I still think we should write a very short tutorial on how to use the `json_configurations`, or add it as part of the "Output" page maybe... I think that `csv_configurations` are also undocumented, so we might as well add that while we are in there.

merschformann:
That is fair, though, we typically import from the root. I wasn't considering the direct import from the module. 

Let me see what I can do with the docs. Technically, not in scope, but I'll give it some time. :relaxed:

merschformann:
@sebastian-quintero : I changed the whole module around to `_serialization`, which makes more sense, I think. The "private" funcs made pylance even warn about them being "unused". It makes sense, as it means to mark funcs as internal to a module, not a package. Making the module "private" somehow works for me. :relaxed:

sebastian-quintero:
Sweet!

merschformann:
Docs added/extended @sebastian-quintero . I think we're ready for the final pass. :blush:

docs/nextmv/tutorials/output.md
sebastian-quintero, line 1: Amazing, thank you for the contributions

sebastian-quintero (approved):
Thank you! LGTM

PR #107, opened 2025-06-11
Prevents numerical overflows during polling

merschformann:
# Description

This PR addresses a numerical overflow (bug) that happens when polling with long timeouts. Furthermore, we change the defaults for polling to not have a timeout or maximum number of polls. This avoids running into unexpected exceptions when polling long runs.

## Changes

- Prevents an overflow that occurs when polling for very long times.
- Remove timeout and max tries from polling defaults.
- Improve test times by allowing to override delay function.

Resolves ENG-6121, ENG-6123

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -184,9 +184,9 @@ class PollingOptions:
     """
     max_delay: float = 20
     """Maximum delay to use between polls, in seconds."""
-    max_duration: float = 300
+    max_duration: float = -1
     """Maximum duration of the polling strategy, in seconds."""
-    max_tries: int = 100
+    max_tries: int = -1
     """Maximum number of tries to use."""
     jitter: float = 1
     """
@@ -3258,7 +3258,13 @@ def poll(  # noqa: C901
 
     # Begin the polling process.
     max_reached = False
-    for ix in range(polling_options.max_tries):
+    ix = 0
+    while True:
+        # Check if we reached the maximum number of tries. Break if so.
+        if ix >= polling_options.max_tries and polling_options.max_tries >= 0:
+            break
+        ix += 1
+
         # Check is we should stop polling according to the stop callback.
         if polling_options.stop is not None and polling_options.stop():
             stopped = True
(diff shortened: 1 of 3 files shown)


phil-nextmv (approved):

nextmv/nextmv/cloud/application.py
```diff
@@ -3174,7 +3174,11 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
         raise ValueError(f"Unknown scenario input type: {scenario.scenario_input.scenario_input_type}")
 
 
-def poll(polling_options: PollingOptions, polling_func: Callable[[], tuple[Any, bool]]) -> Any:
+def poll(  # noqa: C901
+    polling_options: PollingOptions,
+    polling_func: Callable[[], tuple[Any, bool]],
+    __sleep_func: Callable[[float], None] = time.sleep,
```
sebastian-quintero, line 3182: Interesting, I have never seen an argument using the `__` style.

sebastian-quintero (approved):

nextmv/nextmv/cloud/application.py
merschformann: This just felt intuitive to me tbf. Happy to workout a different way. I just think that we need to mock away the sleep functionality as the function can't be tested otherwise. Happy to alternatively inject it via some "global" variable if that feels nicer. :relaxed:

sebastian-quintero: I think it is fine, was just commenting about the arg name using `_ _ `, no action needed or anything, was just commenting.

nextmv/nextmv/cloud/application.py
```diff
@@ -184,9 +184,9 @@ class PollingOptions:
     """
     max_delay: float = 20
     """Maximum delay to use between polls, in seconds."""
-    max_duration: float = 300
+    max_duration: float = -1
     """Maximum duration of the polling strategy, in seconds."""
-    max_tries: int = 100
+    max_tries: int = -1
```
sebastian-quintero: Can you update the docstrings at the top? With the new defaults?

sebastian-quintero (approved):
Can you update the docs please?

nextmv/nextmv/cloud/application.py
merschformann: Oh yes, sorry I forgot.

PR #112, opened 2025-07-03
Add support for `multi-file`

sebastian-quintero:
# Description

This PR adds support for `multi-file` in various ways.

- Introduces the `DataFile` class to be able to read an arbitrary file from some location. Convenience functions are added to read JSON, CSV and utf-8 text files. A list of these `DataFile` objects is given to `load`, to construct the `Input` from those. The `Input.data` in this case would hold a dict, where the keys are the file names and the values are the file contents, which are read according to the instructions of `DataFile`.
- Similarly, introduces the `SolutionFile` class to be able to write some data (a solution) to a file in some location. Convenience functions are also added for this case. These `SolutionFile` objects are given to the `Output`, so that when we write the output, if those files are present then they are written to `outputs`.
- Support for writing statistics and assets to `outputs/statistics` and `outputs/assets` folders.
- Adds support for `MULTI_FILE` and `CSV_ARCHIVE` in the `new_run` and `new_run_with_result` methods. If the run configuration is set to the appropriate format, and the `dir_path` is specified, we now tar those files and use the presigned URLs to upload the tarball. This was tested and it works.

## Example

Here is an example of using `multi-file`. It takes the ortools knapsack csv archive app, and adapts it for multi-file. In the example, CSV inputs are consumed and JSON outputs are produced.

```python
import time

from ortools.linear_solver import pywraplp

import nextmv

# Status of the solver after optimizing.
STATUS = {
    pywraplp.Solver.FEASIBLE: "suboptimal",
    pywraplp.Solver.INFEASIBLE: "infeasible",
    pywraplp.Solver.OPTIMAL: "optimal",
    pywraplp.Solver.UNBOUNDED: "unbounded",
    pywraplp.Solver.ABNORMAL: "abnormal",
    pywraplp.Solver.NOT_SOLVED: "not_solved",
    pywraplp.Solver.MODEL_INVALID: "model_invalid",
}


def main() -> None:
    """Entry point for the program."""

    options = nextmv.Options(
        nextmv.Option("input", str, "", "Path to input folder.", False),
        nextmv.Option("output", str, "", "Path to output folder.", False),
        nextmv.Option("duration", int, 30, "Max runtime duration (in seconds).", False),
        nextmv.Option("provider", str, "SCIP", "Solver provider.", False),
    )

    input = nextmv.load(
        input_format=nextmv.InputFormat.MULTI_FILE,
        options=options,
        path=options.input,
        data_files=[
            nextmv.csv_data_file("items"),
            nextmv.csv_data_file("weight_capacity"),
        ],
    )

    nextmv.log("Solving knapsack problem:")
    nextmv.log(f"  - items: {len(input.data.get('items', []))}")
    nextmv.log(f"  - capacity: {input.data.get('weight_capacity', 0)}")

    model = DecisionModel()
    output = model.solve(input)
    nextmv.write(output, path=options.output)


class DecisionModel(nextmv.Model):
    def solve(self, input: nextmv.Input) -> nextmv.Output:
        """Solves the given problem and returns the solution."""

        start_time = time.time()
        nextmv.redirect_stdout()  # Solver chatter is logged to stderr.

        # Creates the solver.
        solver = pywraplp.Solver.CreateSolver(input.options.provider)
        solver.SetTimeLimit(input.options.duration * 1000)

        # Initializes the linear sums.
        weights = 0.0
        values = 0.0

        # Creates the decision variables and adds them to the linear sums.
        items = []
        for item in input.data["items.csv"]:
            item_variable = solver.IntVar(0, 1, item["id"])
            items.append({"item": item, "variable": item_variable})
            weights += item_variable * int(item["weight"])
            values += item_variable * int(item["value"])

        # This constraint ensures the weight capacity of the knapsack will not be
        # exceeded.
        capacity = int(input.data["weight_capacity.csv"][0]["weight_capacity"])  # Read as a CSV.
        solver.Add(weights <= capacity)

        # Sets the objective function: maximize the value of the chosen items.
        solver.Maximize(values)

        # Solves the problem.
        status = solver.Solve()

        # Determines which items were chosen.
        chosen_items = [item["item"] for item in items if item["variable"].solution_value() > 0.9]

        statistics = nextmv.Statistics(
            run=nextmv.RunStatistics(duration=time.time() - start_time),
            result=nextmv.ResultStatistics(
                duration=solver.WallTime() / 1000,
                value=solver.Objective().Value(),
                custom={
                    "status": STATUS.get(status, "unknown"),
                    "variables": solver.NumVariables(),
                    "constraints": solver.NumConstraints(),
                },
            ),
        )

        return nextmv.Output(
            output_format=nextmv.OutputFormat.MULTI_FILE,
            options=input.options,
            statistics=statistics,
            solution_files=[nextmv.json_solution_file("solution", chosen_items)],
        )


if __name__ == "__main__":
    main()
```


nextmv/nextmv/input.py
```diff
@@ -528,6 +825,73 @@ def _load_archive(
...
+
+        dir_path = "inputs"
+        if path is not None and path != "":
+            if not os.path.isdir(path):
+                raise ValueError(f"path {path} is not a directory")
+
+            dir_path = path
+
+        if not os.path.isdir(dir_path):
+            raise ValueError(f'expected input directoy "{dir_path}" to exist as a default location')
```
merschformann, line 922:
```suggestion
            raise ValueError(f'expected input directory "{dir_path}" to exist as a default location')
```

Small typo. Detected by Copilot.

nextmv/requirements.txt
```diff
@@ -6,6 +6,8 @@ requests>=2.31.0
 ruff>=0.1.7
 twine>=4.0.2
 mlflow>=2.19.0
+pandas>=2.2.3
+openpyxl>=3.1.5
```
merschformann: Just confirming: these are only dev requirements, right? I still trip a bit over the naming of the file. :blush: 
All good though, I think.

merschformann (commented):
Looks good, thanks for all the work! :hugs: 

I was just playing with it, and I suppose one way of using it is like this, right?

```python
input = nextmv.load(
    options=options,
    path=options.input,
    data_files=[
        nextmv.csv_data_file("items"),
        nextmv.csv_data_file("capacities"),
    ],
    input_format=nextmv.InputFormat.MULTI_FILE,
)
```

It's nice that it doesn't **have to** have the the extension in the file reference (both work though which is awesome :raised_hands: ). I noticed that the key I get returned in the `input.data` has the extension in both cases. I think it would be preferable that we use whatever the user specifies as the key. :blush:

nextmv/requirements.txt
sebastian-quintero: Yes! These are dev requirements. I will change to use optional dependencies like `nextpipe` in the pyproject.

nextmv/nextmv/input.py
sebastian-quintero: Thank you AI.

nextmv/requirements.txt
merschformann: All good, I was only concerned for a second that we now incur pandas as a dependency for all `nextmv` installations (which increases size substantially). But just my bad in not remembering how we handle it in this repo.

sebastian-quintero:
> Looks good, thanks for all the work! 🤗
> 
> I was just playing with it, and I suppose one way of using it is like this, right?
> 
> ```python
> input = nextmv.load(
>     options=options,
>     path=options.input,
>     data_files=[
>         nextmv.csv_data_file("items"),
>         nextmv.csv_data_file("capacities"),
>     ],
>     input_format=nextmv.InputFormat.MULTI_FILE,
> )
> ```
> 
> It's nice that it doesn't **have to** have the the extension in the file reference (both work though which is awesome 🙌 ). I noticed that the key I get returned in the `input.data` has the extension in both cases. I think it would be preferable that we use whatever the user specifies as the key. 😊

The issue here is what if you have `items.json` and `items.csv`, independently of how you use the helper funcs or if you implement the class yourself, the files have the same name... What I am thinking here is to introduce a new field in the Class that is the "key" that the data will have in the output. It would be optional, and if the user doesn't specify it, then the key is the file name. If the user specifies it, then the key is waht they specify.

I'll do that.

nextmv/requirements.txt
sebastian-quintero: For a hot second it was going to be a dependency, to handle Excel, but I came up with the `DataFile` and `SolutionFile` abstraction where people can pass their own `Callable` and then it goes away, after all, we don't know how people want to parse their Excel files.

sebastian-quintero: I changed it to the pyproject.

merschformann: Yeah, :100: percent!

sebastian-quintero:
@merschformann I added support for custom keys when loading the data from the files and I added the dev dependencies to the pyproject, like in nextpipe, for the three packages. Let me know if this looks good.

merschformann (approved):
Awesome, thank you @sebastian-quintero ! :hugs:

PR #114, opened 2025-07-10
Fixes application upload logic

sebastian-quintero:
Corrects a boolean return value in the upload ID check and adds a check to prevent empty tar files being sent, addressing potential errors in application deployment.


nextmv/nextmv/cloud/client.py
```diff
@@ -397,7 +397,7 @@ def upload_to_presigned_url(
 
         if upload_data is not None:
             kwargs["data"] = upload_data
-        elif tar_file is not None:
+        elif tar_file is not None and tar_file != "":
```
merschformann, line 400:
```suggestion
        elif tar_file:
```

Isn't it equivalent to this?

Almost I suppose. Actually, potentially better to be more explicit. :relaxed: 
So, feel free to leave it like it is.

sebastian-quintero: Yes I prefer to be explicit... I'll leave it as is...

merschformann (approved):

PR #117, opened 2025-07-11
Release is now based on PRs

sebastian-quintero:
# Description

Changes the release process to be automated after a PR is merged. To release, the new version has to be proposed with a PR.

Please see the new `RELEASE.md` file for details.

Removes integration to TestPyPI to simplify.


nextmv-gurobipy/nextmv_gurobipy/__about__.py
sebastian-quintero, line 1: After the merge, it is probable that the release process fails becase these versions exist already. The idea is that I can create a stable release after this PR, by opening another PR and setting the versions to the correct number, which would be a stable version, not a pre-release version.

merschformann (commented):
This is getting quite complicated. I think the idea is right, we need to manually bump the version to get a release out as we can't auto-push changes anymore. However, let's then fully commit to that new way and remove the input parameters, bot setup, etc. ... basically all that is here only for the old way (which works on non-main/develop branch, but the new way also works on those branches, right?). :thinking:

sebastian-quintero:
Need to create a reusable workflow for this, before it gets out of hand and have to copy and paste everywhere.

merschformann (approved):

PR #121, opened 2025-08-04
Refactor batch exp creation with instance ids

sebastian-quintero:
# Description

The backend will soon drop support for a list of instance ids when creating a batch experiment. When using `instance_ids` in batch experiment creation, we now emit a deprecation warning and instead of using them, we translate them to `runs`, with help of the input set.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -45,6 +45,7 @@ from nextmv.cloud.batch_experiment import (
     BatchExperimentInformation,
     BatchExperimentMetadata,
     BatchExperimentRun,
+    to_runs,
 )
 from nextmv.cloud.client import Client, get_size
 from nextmv.cloud.input_set import InputSet, ManagedInput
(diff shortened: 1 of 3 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -1207,7 +1219,14 @@ def new_batch_experiment(
         if input_set_id is not None:
             payload["input_set_id"] = input_set_id
         if instance_ids is not None:
-            payload["instance_ids"] = instance_ids
+            deprecated(
+                name="new_batch_experiment.instance_ids",
+                reason="using argument `instance_ids` is deprecated, use `runs` instead",
+            )
```
merschformann, line 1225: I don't think we need to deprecate comparing instances here, since the Python SDK means to provide convenience functionality on top of the bare backend API.

I don't have strong feelings on it though, so, I won't block. :relaxed:

merschformann (approved):
Thanks for the update! :hugs:

nextmv/nextmv/cloud/application.py
sebastian-quintero: If it will be deprecated at some point in the API, I would prefer to keep the concepts of runs here as well, for consistency. Thank you for pointing it out, though.

PR #123, opened 2025-08-08
Introduce application initialization

sebastian-quintero:
# Description

Initialize a new application, locally, with `Application.initialize`. This will create a new dir in the desired destination with an out-of-the-box structure.


nextmv/nextmv/cloud/manifest.py
```diff
@@ -649,9 +654,10 @@ def from_options(cls, options: Options, validation: OptionsEnforcement = None) -
         return cls(
             strict=validation.strict if validation else False,
             validation=ManifestValidation(enforce="all" if validation and validation.validation_enforce else "none"),
-            items=items
+            items=items,
```
merschformann, line 657: It appears that this file didn't get proper formatting before when the missing comma was introduced. We should make sure that formatting (with ruff) is applied by all of us. :thinking:

nextmv/nextmv/default_app/src/main.py
```diff
@@ -0,0 +1,36 @@
...
+)
+
+##### Insert model here
+
+# Print logs that render in the run view in Nextmv Console.
+message = f"Hello, {name}"
+nextmv.log(message)
+
+if options.details:
+    detail = "You are", {input.data["distance"]}, " million km from the sun"
```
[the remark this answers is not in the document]
merschformann: Yeah, good catch.

nextmv/nextmv/default_app/requirements.txt
```diff
@@ -0,0 +1,2 @@
+nextmv>=0.29.3
+plotly>=6.2.0
```
merschformann, line 2: Feels like a bit of a heavy dependency to bring in for the init app. Getting this:

```bash
📦 Packaged application (18.06 MiB, 4184 files).
```

But I don't have a strong opinion on it. It does show the use of visuals (one of them). geojson would be cheaper dependency wise, but :shrug:

nextmv/nextmv/default_app/README.md
```diff
@@ -0,0 +1,17 @@
+# Nextmv Application
+
+This is the basic structure of a Nextmv application.
+
+```text
+├── app.yaml
+├── README.md
+├── requirements.txt
+└── src
```
merschformann, line 9: This app structure doesn't work without the _main.py_, right? Am I missing something? Or, is this already preparation work intended to be used with custom entry point control? :thinking:

merschformann (approved):
I have some questions and there is the comment from Copilot that should get addressed. I might be missing context here, so, not blocking.

phil-nextmv (approved):

nextmv/nextmv/cloud/manifest.py
sebastian-quintero: I am personally using ruff, so maybe one of those cases where it passes linting but the formatter would rather have a comma 🤷🏻‍♂️

nextmv/nextmv/default_app/requirements.txt
sebastian-quintero: I'll merge this one and in the next pass I'll just change to geojson. This was borrowed from community apps.

nextmv/nextmv/default_app/README.md
sebastian-quintero: Yes, this is preparation work for custom entry point. The idea is that `src/main.py` is the entrypoint.

nextmv/nextmv/cloud/manifest.py
merschformann: It wasn't you, ruff always adds it. :blush:

PR #128, opened 2025-09-02
All fields on update methods are optional

sebastian-quintero:
# Description

Modifies the `Application.update_xyz` methods to make all the fields optional. Where applicable, the entity is fetched from the API and only the specified fields are updated. This is done to "gracefully" handle `PUT` methods.

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -2755,7 +2755,7 @@ class Application:
     def update_instance(
         self,
         id: str,
-        name: str,
+        name: Optional[str] = None,
         version_id: Optional[str] = None,
         description: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


merschformann (approved):
As mentioned internally, I think, making it more forward-compatible robust would be icing on the cake. However, no reason to block the fix. Thanks for the work @sebastian-quintero ! :hugs:

sebastian-quintero:
> As mentioned internally, I think, making it more forward-compatible robust would be icing on the cake. However, no reason to block the fix. Thanks for the work @sebastian-quintero ! 🤗

I think it would be a greater change to support this, as I am currently using the methods the SDK already has implemented. If we see those APIs changing frequently (which I doubt, and hence my push not to do it), then I am on board with implementing a more open-ended solution. Thanks.

PR #129, opened 2025-09-03
Add option formatting support to manifest

dirkschumacher:
Adds option formatting support to the app manifest, allowing users to specify how options are transformed into command line arguments.

### DIFF
--- a/nextmv/nextmv/cloud/manifest.py
+++ b/nextmv/nextmv/cloud/manifest.py
@@ -601,6 +601,10 @@ class ManifestOptions(BaseModel):
         is a parameter that configures the decision model.
     validation: Optional[ManifestValidation], default=None
         Optional validation rules for all options.
+    format: Optional[list[str]], default=None
+        A list of strings that define how options are transformed into command
+        line arguments. Use `{{name}}` to refer to the option name and
+        `{{value}}` to refer to the option value.
 
 
     Examples
@@ -629,9 +633,21 @@ class ManifestOptions(BaseModel):
 
     An option is a parameter that configures the decision model.
     """
+    format: Optional[list[str]] = None
+    """A list of strings that define how options are transformed into command line arguments.
+
+    Use `{{name}}` to refer to the option name and `{{value}}` to refer to the option value.
+    For example, `["-{{name}}", "{{value}}"]` will transform an option named `max_vehicles`
+    with a value of `10` into the command line argument `-max_vehicles 10`.
+    """
 
     @classmethod
-    def from_options(cls, options: Options, validation: OptionsEnforcement = None) -> "ManifestOptions":
+    def from_options(
+        cls,
+        options: Options,
+        validation: OptionsEnforcement = None,
+        format: Optional[list[str]] = None,
+        ) -> "ManifestOptions":
         """
         Create a `ManifestOptions` from a `nextmv.Options`.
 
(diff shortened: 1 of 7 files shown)


merschformann (commented):
Looks good. Can you add a PR description? It's used for the auto-generated changelog. :blush:

nextmv/nextmv/cloud/manifest.py
```diff
@@ -629,9 +633,21 @@ class ManifestOptions(BaseModel):
...
+    """
 
     @classmethod
-    def from_options(cls, options: Options, validation: OptionsEnforcement = None) -> "ManifestOptions":
+    def from_options(
+        cls,
+        options: Options,
+        validation: OptionsEnforcement = None,
+        format: Optional[list[str]] = None,
+        ) -> "ManifestOptions":
```
sebastian-quintero: Indenting seems off here. I think unindent to the left.

sebastian-quintero (commented):
I don't quite follow why this is necessary and no changes to the `options.py` is warranted... In there, `Options` is the class that has the logic for parsing options and reading them from CLI args or env vars. Seems to me like customizing the format of how they are recognized as CLI args would warrant a change in `options.py`, `Options` class as well.

sebastian-quintero:
@dirkschumacher it would be good to also modify the Python docs with this PR. They live under `./docs`.

nextmv/nextmv/cloud/manifest.py
merschformann: FYI @dirkschumacher : we use ruff as the linter. 
Also, I missed a change in `options.py`. That should indeed not be necessary. :relaxed:

sebastian-quintero: Why wouldn't it be necessary, what am I missing 😅 ?

merschformann: I think we are confusing the custom formatting of the options part here that is necessary for apps not adhering to our convention. Our own options adhere to our convention though, so, not necessary to let them deal with anything. At least, not in the current scope. 
We always need to carefully distinguish `nextmv` and `nextmv.cloud` namespaces here. The latter is applicable to non-python apps, etc.

sebastian-quintero: Ok, so in my mind I don't necessarily try to distinguish `nextmv` and `nextmv.cloud` because the idea (at least for me) is that they always play nicely with each other and not drift apart (too much, at least). Ok, so with what you say:

> necessary for apps not adhering to our convention

Then it makes sense, for as you put it: apps built with the SDK whould always follow convention. Thank you for clarifying.

dirkschumacher:
> I don't quite follow why this is necessary and no changes to the `options.py` is warranted... In there, `Options` is the class that has the logic for parsing options and reading them from CLI args or env vars. Seems to me like customizing the format of how they are recognized as CLI args would warrant a change in `options.py`, `Options` class as well.

My understanding is that the standard options use the default command line argument formatting, so there is no need to modify the formatting template when using these options. Or am I missing something?

merschformann (approved):
I think the docs are already covered sufficiently for the scope of the work by the code comments in manifest.py.
If we want tutorials or similar, we should add something coherent as it would involve other parts of the `app.yaml` that are also not documented separately (like option validation, ui options, etc.).

Anyway, to me this is ready to go. :relaxed:

PR #132, opened 2025-09-13
The Python SDK community edition

sebastian-quintero:
# Description

This PR finishes the community edition with the local package.


docs/nextmv/reference/cloud/safe.md
```diff
@@ -0,0 +1,5 @@
+# Safe Module
+
+This section documents the sagfe components of the Nextmv Cloud API.
```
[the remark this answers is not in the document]
sebastian-quintero: Done

nextmv/nextmv/cloud/application.py
```diff
@@ -3594,6 +3705,74 @@ def __upload_url_required(
...
+        json_configurations: Optional[dict[str, Any]] = None,
+    ) -> dict[str, str]:
+        """
+        Auxiliary function to extract the options that will be sent to the
+        application for execution.
+        """
+
+        options_dict = {}
+        if isinstance(input, Input) and input.options is not None:
+            options_dict = input.options.to_dict_cloud()
```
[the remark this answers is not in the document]
sebastian-quintero: Thanks

nextmv/nextmv/cloud/application.py
```diff
@@ -2677,13 +2387,26 @@ def track_run(self, tracked_run: TrackedRun, instance_id: Optional[str] = None)
         information about a run in Nextmv is useful for things like
         experimenting and testing.
 
+        Please read the documentation on the `TrackedRun` class carefully, as
+        there are important considerations to take into account when using this
+        method. For example, if you intend to upload JSON input/output, use the
+        `input`/`output` attributes of the `TrackedRun` class. On the other
+        hand, if you intend to track files-based input/output, use the
+        `input_dir_path`/`output_dir_path` attributes of the `TrackedRun`
+        class.
```
merschformann, line 2732: I don't think non-json tracked runs are supported yet. It might be on the horizon though. Is this aligned with a future extension?

nextmv/nextmv/local/application.py
```diff
@@ -0,0 +1,1145 @@
...
+        Syncs a local run to a Nextmv Cloud target application. Returns True if
+        the run was synced, False if it was skipped (already synced).
+        """
+
+        if verbose:
+            log(f"🔄 Syncing local run `{run_id}`... ")
+
+        # For files-based runs, the result files are written by default to ".".
+        # Avoid this using a dedicated temp dir.
+        run_result = self.run_result(run_id, output_dir_path=temp_dir)
```
merschformann, line 968: How do we make sure that the run-ids do not collide? The run-id is specified by the user, right? :thinking:

nextmv/nextmv/local/application.py
```diff
@@ -0,0 +1,1145 @@
...
+            tracked_run.output_dir_path = os.path.join(run_dir, OUTPUTS_KEY, SOLUTIONS_KEY)
+
+        # Actually sync the run by tracking it remotely on Nextmv Cloud.
+        configuration = RunConfiguration(
+            format=Format(
+                format_input=run_result.metadata.format.format_input,
+                format_output=run_result.metadata.format.format_output,
+            ),
+        )
+        tracked_id = target.track_run(
```
merschformann, line 1044: When syncing a run with problematic input (binary files, etc.) or output (too large, etc.) the sync will simply fail, right? Will it try to re-sync and fail every time the user tries to sync all runs? :thinking:

nextmv/nextmv/local/executor.py
```diff
@@ -0,0 +1,565 @@
...
+            with open(info_file, "r+") as f:
+                info = json.load(f)
+                info["metadata"]["status_v2"] = "running"
+                f.seek(0)
+                json.dump(info, f, indent=2)
+                f.truncate()
+
+            # Start a Python subprocess to execute the entrypoint. For now, we are
+            # supporting a Python-first experience, so we are not summoning
+            # applications that are not Python-based.
```
merschformann, line 149: Thanks for clarifying, I was wondering about that. :relaxed:

nextmv/nextmv/local/executor.py
```diff
@@ -0,0 +1,565 @@
...
+    logs_dir = os.path.join(run_dir, LOGS_KEY)
+    os.makedirs(logs_dir, exist_ok=True)
+
+    # The complete execution is wrapped to capture any errors.
+    try:
+        # Create a temp dir, and copy the entire src there, to have a transient
+        # place to work from, and be cleaned up afterwards.
+        with tempfile.TemporaryDirectory() as temp_dir:
+            temp_src = os.path.join(temp_dir, "src")
+            shutil.copytree(src, temp_src, ignore=shutil.ignore_patterns(".nextmv"))
```
merschformann, line 126: We are not copying / handling the requirements from the manifest / requirements.txt file, right? :thinking: 
Is this a plan in the future, or, do we simply use the local installation for the local experience?

nextmv/nextmv/local/executor.py
```diff
@@ -0,0 +1,565 @@
...
+    run_dir : str
+        The path to the run directory.
+    result : subprocess.CompletedProcess[str]
+        The result of the subprocess run.
+    """
+
+    logs_dir = os.path.join(run_dir, LOGS_KEY)
+    os.makedirs(logs_dir, exist_ok=True)
+    with open(os.path.join(logs_dir, LOGS_FILE), "w") as f:
+        f.write(result.stderr)
```
merschformann: I think we need a bit more complex stdout/stderr handling logic. For `multi-file` we consider stdout+stderr as the log-stream. So, we need to have them both merged into one log stream when capturing. I don't think we can do this post run, right? :thinking:

nextmv/nextmv/local/executor.py
```diff
@@ -0,0 +1,565 @@
...
+    ----------
+    temp_run_outputs_dir : str
+        The path to the temporary outputs directory.
+    outputs_dir : str
+        The path to the outputs directory in the run directory.
+    stdout_output : dict[str, Any]
+        The stdout output of the run, parsed as a dictionary.
+    """
+
+    stats_src = os.path.join(temp_run_outputs_dir, STATISTICS_KEY)
```
merschformann: I think in case of `multi-file` statistics are just one file (which location can be configured now - live in prod). So, I don't think a directory works here, right?
For `csv-archive` and `json` just checking the `statistics` key makes sense. I am not sure whether ignoring the content type of the run is stable enough, but might be fine.

nextmv/nextmv/local/geojson_handler.py
```diff
@@ -0,0 +1,323 @@
...
+        if "properties" not in first_feature or not first_feature["properties"]:
+            return tooltip_fields, popup_fields
+
+        available_fields = list(first_feature["properties"].keys())
+        # Prioritize common field names for tooltip/popup
+        priority_fields = ["name", "title", "label", "id", "popupContent", "description"]
+
+        for field in priority_fields:
+            if field in available_fields:
+                tooltip_fields.append(field)
```
merschformann, line 256: I like the addition of tooltips and popups, but I think we should align with what Console does. I think, Console uses specific fields/keys for different intentions:

- "properties.name": Used for filtering (and the popup title?)
- "properties.style": A _map_ that understands "color", "fillColor", etc.
- "properties.metadata": A _list_ of key-value pairs like these: `{"key": "Vehicle ID", "value": "v1"}`

We should sync on this again though. I have an example of this, I think and we can check with David.

nextmv/nextmv/local/runner.py
```diff
@@ -0,0 +1,312 @@
...
+    args = [sys.executable, "executor.py"]
+    process = subprocess.Popen(
+        args,
+        env=os.environ,
+        text=True,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+        cwd=os.path.dirname(__file__),
+        start_new_session=True,  # Detach from parent process
```
merschformann, line 136: Is there a way to interact with an ongoing run(s)? Maybe not right now, but in future? :thinking: 
I can easily imagine spawning a number of runs (maybe even accidentally) and then not knowing when they are done, what their status is, etc.

nextmv/tests/local/__init__.py
merschformann, line 1: Should this have explicit exports? :thinking:

nextmv/nextmv/run.py
merschformann, line 1: Will the types of run be dual exported (from cloud AND from local)? :thinking: 
I hope this is all not causing too much confusion later. Now that we are adding yet another role for `nextmv` as a package.

merschformann (approved):
I am running out of review time. :sweat_smile: 

Looks good. I do have some thoughts in general and left some comments. I think it would be good to release small things first before unleashing this kraken. :blush: 
I don't want to block this though. Happy to sync internally. 

Great work! Very clean! :hugs:

nextmv/nextmv/cloud/application.py
sebastian-quintero: Baking in support to be future-proof.

nextmv/nextmv/local/application.py
sebastian-quintero: The run ID is not user generated

nextmv/nextmv/local/application.py
sebastian-quintero: Correct, if syncing a run is problematic from `cloud` side, then syncing that run fails.

nextmv/nextmv/local/executor.py
sebastian-quintero: We use the local installation given by `sys.executable`, so we don't care much about the requirements at the moment.

nextmv/nextmv/local/executor.py
sebastian-quintero: Fixed to stream stdout and stderr to the log file for multi-file

nextmv/nextmv/local/executor.py
sebastian-quintero: Adapted for multi-file to use the configurations in the manifest if they are defined or defaulting to this when not.

nextmv/nextmv/local/runner.py
sebastian-quintero: Yes, when a run is ongoing you can poll for results and the normal stuff.

nextmv/tests/local/__init__.py
sebastian-quintero: I avoided it to prevent importing optional deps.

nextmv/nextmv/run.py
sebastian-quintero: I am still exporting stuff from `cloud` for backwards compatibility but in the future the types established as part of `nextmv` should only be exported from `nextmv`.

PR #141, opened 2025-09-22
Add `format` field to `Metadata`

cnpryer:
# Description

Adds missing format field in metadata.

## Changes

- Adds format field in run metadata.


merschformann (approved):
Awesome! Thank you for noticing and adding! :hugs:

PR #143, opened 2025-10-02
Feature/eng 5549 support ensemble runs via python sdk

cowanwalls:
# Description

Adds ensemble definition support.

## Changes

- `nextmv/nextmv/cloud/ensemble.py`
    - Ensemble definition types added
- `nextmv/nextmv/cloud/__init__.py`
    - Ensemble definition types exported to cloud
        - Should we consolidate `ToleranceType` in acceptance tests and Ensemble runs to a single type? Maybe via the addition of a `statatics.py` file in `nextmv/nextmv/cloud`?
- `nextmv/nextmv/cloud/application.py`
    - Adds ensemble management functions
    - `new_version` now defaults the `name` field to the value of `id` if `name` is `None` (fixes unexpected 404)
    
## Testing

Manually tested these changes.


nextmv/nextmv/cloud/application.py
```diff
@@ -551,6 +556,30 @@ def delete_secrets_collection(self, secrets_collection_id: str) -> None:
             endpoint=f"{self.endpoint}/secrets/{secrets_collection_id}",
         )
 
+    def delete_ensemble_definition(self, ensemble_definition_id: str) -> None:
```
sebastian-quintero: Nit: we try to organize the methods in this class alphabetically, so this one should be above `delete_scenario_test`.

nextmv/nextmv/cloud/application.py
```diff
@@ -551,6 +556,30 @@ def delete_secrets_collection(self, secrets_collection_id: str) -> None:
...
+        """
+        Delete a secrets collection.
```
sebastian-quintero: This is not accurate.

nextmv/nextmv/cloud/application.py
```diff
@@ -683,6 +712,39 @@ def instance_exists(self, instance_id: str) -> bool:
                 return False
             raise e
 
+    def ensemble_definition(self, ensemble_definition_id: str) -> EnsembleDefinition:
```
sebastian-quintero: Nit: same comment about ordering methods alphabetically inside the class.

nextmv/nextmv/cloud/application.py
```diff
@@ -796,6 +858,36 @@ def list_instances(self) -> list[Instance]:
 
         return [Instance.from_dict(instance) for instance in response.json()]
 
+    def list_ensemble_definitions(self) -> list[EnsembleDefinition]:
```
sebastian-quintero: Nit: same comment here about ordering alphabetically.

nextmv/nextmv/cloud/application.py
```diff
@@ -796,6 +858,36 @@ def list_instances(self) -> list[Instance]:
...
+        """
+        List all ensemble_definitions.
+
+        Returns
+        -------
+        list['EnsembleDefinition']
```
sebastian-quintero:
```suggestion
        list[EnsembleDefinition]
```

I think in this case backticks are not needed.

nextmv/nextmv/cloud/application.py
```diff
@@ -1280,6 +1372,53 @@ def new_batch_experiment_with_result(
...
+            ID of the ensemble defintion.
+        name: Optional[str]
+            Name of the ensemble definition.
+        description: Optional[str]
+            Description of the ensemble definition.
+        run_groups: Optional[list[RunGroup]]
+            Information to facilitate the execution of child runs.
+        rules: Optional[list[EvaluationRule]]
+            Information to facilitate the selection of
+            a result for the ensemble run from child runs.
```
sebastian-quintero: Parameters should be in the same order they are expected in the function arguments.

nextmv/nextmv/cloud/application.py
```diff
@@ -1280,6 +1372,53 @@ def new_batch_experiment_with_result(
...
+            "id": id,
+            "run_groups": [run_group.to_dict() for run_group in run_groups],
+            "rules": [rule.to_dict() for rule in rules],
+            "name": name,
+            "description": description,
+        }
+
+        response = self.client.request(
+            method="POST",
+            endpoint=f"{self.endpoint}/ensembles",
```
sebastian-quintero: Nit: consider defining `ensembles` as an attribute of the class, similar to `experiments_endpoint`:

```python
experiments_endpoint: str = "{base}/experiments"
```

That way, you can use `{self.ensembles_endpoint}` as opposed to `{self.endpoint}/ensembles` here and in other places.

nextmv/nextmv/cloud/application.py
```diff
@@ -2963,6 +3103,54 @@ def update_instance(
 
         return Instance.from_dict(response.json())
 
+    def update_ensemble_definition(
```
sebastian-quintero, line 3124: Nit: same comment about alphabetical order.

nextmv/nextmv/cloud/__init__.py
```diff
@@ -63,6 +63,12 @@
 from .batch_experiment import ExperimentStatus as ExperimentStatus
 from .client import Client as Client
 from .client import get_size as get_size
+from .ensemble import EnsembleDefinition as EnsembleDefinition
```
sebastian-quintero, line 67: Strange, this should be ok.

sebastian-quintero (changes requested):
This is looking really good, congratulations! Aside from the specific comments, please add the new `ensemble.py` to the docs reference. For that there are two changes you need to make:

1. In the `mkdocs.yml` file in the root of the repo, add this reference in the appropriate location, which should be under `- client.py: nextmv/reference/cloud/client.md`, to keep the alphabetical order. You can simply copy that line and change `client` for `ensemble`.
2. Because you are referencing a non-existent `ensemble.md` file, you need to create it. Add it to `docs/nextmv/reference/cloud`. You can copy any of the existing files there, rename it, and modify the content to point to the correct reference `::: nextmv.nextmv.cloud.ensemble`.

Basically, whenever we add _new_ files to the SDK, we need to add them to the technical reference of docs as well.

sebastian-quintero:
@cowanwalls You should `pip install ruff` locally and run `ruff check .` to make sure that the linting passes locally....

sebastian-quintero (approved):
My only leftover request is for this new file to be exposed as reference. Look at the cloud section in the mkdocs.yml index, you should see how to add the file as .md as well. This for docs to be up to date. Would really be awesome to have a dedicated page of a tutorial to handle ensemble definitions and runs in general.

PR #144, opened 2025-10-02
Returns the runs of a batch experiment, adds the `list_runs` methods for `Application` in `local` and `cloud`

sebastian-quintero:
# Description

When using the `batch_experiment` methods to retrieve the results of a batch experiment, a new `runs` property was added to `BatchExperiment`, to access the runs of that batch exp. Behind the covers, the separate `/runs` endpoint is used.

Because a new `Run` class was introduced to capture the information about a run, iI took the chance to introduce the `list_runs` method, which uses that exact class. This method is added to both applications: in `cloud` and `local`.

### DIFF
--- a/nextmv/nextmv/local/local.py
+++ b/nextmv/nextmv/local/local.py
@@ -84,8 +84,9 @@ def calculate_files_size(run_dir: str, run_id: str, dir_path: str, metadata_key:
         for f in filenames:
             fp = os.path.join(dirpath, f)
             # Skip if it is a symbolic link
-            if not os.path.islink(fp):
-                total_size += os.path.getsize(fp)
+            if os.path.islink(fp):
+                continue
+            total_size += os.path.getsize(fp)
 
     info_file = os.path.join(run_dir, f"{run_id}.json")
     with open(info_file, "r+") as f:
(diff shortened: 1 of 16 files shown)


nextmv/nextmv/local/local.py
```diff
@@ -0,0 +1,96 @@
...
+        metadata.
+    """
+
+    total_size = 0
+    for dirpath, _, filenames in os.walk(dir_path):
+        for f in filenames:
+            fp = os.path.join(dirpath, f)
+            # Skip if it is a symbolic link
+            if not os.path.islink(fp):
+                total_size += os.path.getsize(fp)
```
merschformann:
```suggestion
            # Skip if it is a symbolic link
            if os.path.islink(fp):
                continue
            total_size += os.path.getsize(fp)
```

**Very** nitpicky, but my inner monk asks me to align code with comment here (or other way around). 😅

merschformann (approved):
Tough to spot the actual changes among some of the restructuring, but looks all good to me. Thanks for the work! 🤗

nextmv/nextmv/local/local.py
sebastian-quintero: Lol all good! Thanks for spotting

PR #150, opened 2025-10-20
Fixes input dir path validation

merschformann:
# Description

Fixes the input validation being too strict on multi-file apps, which have their input/output format specified in the app's manifest (instead of explicitly on the run).

## Changes

- Removes strict input dir path validation to enable `multi-file` runs driven by app manifest settings.

Resolves ENG-6704

### DIFF
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -3909,6 +3909,8 @@ class Application:
         ):
             return config.format.format_input.input_type
 
+        return None
+
     def __package_inputs(self, dir_path: str) -> str:
         """
         This is an auxiliary function for packaging the inputs found in the
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -3879,7 +3879,6 @@ class Application:
             return
 
         # Validate that the input directory path is provided when explicitly required.
-        input_type = configuration.format.format_input.input_type
         dir_types = (InputFormat.MULTI_FILE, InputFormat.CSV_ARCHIVE)
         if input_type in dir_types and not input_dir_path:
             raise ValueError(
--- a/nextmv/nextmv/cloud/application.py
+++ b/nextmv/nextmv/cloud/application.py
@@ -3867,70 +3867,47 @@ class Application:
     def __validate_input_dir_path_and_configuration(
         self,
         input_dir_path: Optional[str],
-        configuration: Optional[Union[RunConfiguration, dict[str, Any]]],
+        configuration: Optional[RunConfiguration],
     ) -> None:
         """
         Auxiliary function to validate the directory path and configuration.
         """
+        input_type = self.__get_input_type(configuration)
 
-        if input_dir_path is None or input_dir_path == "":
+        # If no explicit input type is defined, there is nothing to validate.
+        if input_type is None:
             return
 
-        if configuration is None:
+        # Validate that the input directory path is provided when explicitly required.
+        input_type = configuration.format.format_input.input_type
+        dir_types = (InputFormat.MULTI_FILE, InputFormat.CSV_ARCHIVE)
+        if input_type in dir_types and not input_dir_path:
             raise ValueError(
-                "If dir_path is provided, a RunConfiguration must also be provided.",
+                f"If RunConfiguration.format.format_input.input_type is set to {input_type}, "
+                "then input_dir_path must be provided.",
             )
 
-        config_format = self.__extract_config_format(configuration)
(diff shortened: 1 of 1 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -3867,70 +3867,47 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
...
+                return None
 
-        return config_format.format_input.input_type
+        if (
+            isinstance(config, RunConfiguration)
+            and config.format is not None
+            and config.format.format_input is not None
+            and config.format.format_input.input_type is not None
+        ):
+            return config.format.format_input.input_type
```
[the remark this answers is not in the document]
merschformann: :+1:

nextmv/nextmv/cloud/application.py
```diff
@@ -3867,70 +3867,47 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
...
+        input_type = self.__get_input_type(configuration)
 
-        if input_dir_path is None or input_dir_path == "":
+        # If no explicit input type is defined, there is nothing to validate.
+        if input_type is None:
             return
 
-        if configuration is None:
+        # Validate that the input directory path is provided when explicitly required.
+        input_type = configuration.format.format_input.input_type
```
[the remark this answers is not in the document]
merschformann: :+1:

nextmv/nextmv/cloud/application.py
```diff
@@ -3867,70 +3867,47 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
...
-        if isinstance(config_format, dict):
-            format_input = config_format.get("format_input") or config_format.get("input")
-            if format_input is None:
-                raise ValueError(
-                    "If dir_path is provided, RunConfiguration.format.format_input must also be provided.",
-                )
-
-            if isinstance(format_input, dict):
-                return format_input.get("input_type") or format_input.get("type")
+    def __get_input_type(self, config: Union[RunConfiguration, dict[str, Any]]) -> Union[InputFormat, None]:
```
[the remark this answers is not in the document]
merschformann: Yup.

nextmv/nextmv/cloud/application.py
```diff
@@ -3867,70 +3867,48 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
...
-                return format_input.get("input_type") or format_input.get("type")
+        if config is None:
+            return None
 
-            return getattr(format_input, "input_type", None)
+        if isinstance(config, dict):
+            try:
+                config = RunConfiguration.from_dict(config)
+            except Exception:
+                return None
```
sebastian-quintero: This shouldn't be necessary, I think. If anything I would add a step somewhere that:
- Checks if it is an instance of `RunConfiguration`
- Checks if it is an instance of `dict`
- Raise a `ValueError` exception saying that  `RunConfiguration` is not a valid type.

nextmv/nextmv/cloud/application.py
```diff
@@ -3867,70 +3867,48 @@ def __input_set_for_scenario(self, scenario: Scenario, scenario_id: str) -> Inpu
     def __validate_input_dir_path_and_configuration(
         self,
         input_dir_path: Optional[str],
-        configuration: Optional[Union[RunConfiguration, dict[str, Any]]],
+        configuration: Optional[RunConfiguration],
```
sebastian-quintero: The `__get_input_type` func assumes it might be a dict, so I think the previous type hinting is correct.

nextmv/nextmv/cloud/application.py
sebastian-quintero: Well, maybe what I said above is a first step, then I guess the second step, in case it is a `RunConfiguration`, is what you have here. However, instead of returning `None` I would raise a `ValueError` exception that clearly states that we cannot construct the `RunConfiguration` from the provided `dict`. This means that the dict is wrong.

sebastian-quintero (commented):
I might need further clarification, unless I am missing something.

nextmv/nextmv/cloud/application.py
merschformann: Yeah, good catch. I removed support before, but brought it back after our first chat.
Fixed the type hint. :heart:

nextmv/nextmv/cloud/application.py
merschformann: You mean that it should error instead of ignore when parsing the dict fails. That is probably fair. I'll change it. :blush:

sebastian-quintero (approved):
LGTM

PR #160, opened 2025-11-03
Adds support for app binary execution configuration

merschformann:
# Description

Adds support for defining how to execute the user binary via manifest configuration.

## Changes

* Adds support for new execution configuration via app manifest.

## Preview

Allows setting entrypoint and working directory like this:

```yaml
execution:
  entrypoint: ./app.py
  cwd: ./src/
```

Resolves ENG-6675


nextmv/nextmv/local/executor.py
```diff
@@ -152,7 +152,7 @@ def execute_run(
             # Start a Python subprocess to execute the entrypoint. For now, we are
             # supporting a Python-first experience, so we are not summoning
             # applications that are not Python-based.
-            entrypoint = os.path.join(temp_src, manifest.entrypoint)
+            entrypoint = os.path.join(temp_src, __determine_entrypoint(manifest))
             args = [sys.executable, entrypoint] + options_args(options)
```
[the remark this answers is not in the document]
merschformann: Yeah, fair point. Not sure whether we are immediately supporting it there too, but I took a stab at it.

nextmv/nextmv/cloud/package.py
```diff
@@ -205,8 +205,16 @@ def __find_files(
 def __confirm_mandatory_files(manifest: Manifest, present_files: list[str]) -> None:
     """Confirm that all mandatory files are present in the given list of files."""
 
-    mandatory_files = _MANDATORY_FILES_PER_TYPE[manifest.type]
+    missing_files = []
```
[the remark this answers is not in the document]
merschformann: :+1:

nextmv/nextmv/local/executor.py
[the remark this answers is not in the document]
merschformann: cc @sebastian-quintero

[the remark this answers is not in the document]
sebastian-quintero: Yeah I think the robot 🤖 is right, it is a good idea to support the cwd and entrypoint if defined in the Manifest.

nextmv/nextmv/local/executor.py
```diff
@@ -190,6 +191,35 @@ def execute_run(
             f.truncate()
 
 
+def __determine_entrypoint(manifest: Manifest) -> str:
```
sebastian-quintero: Stupid nit: can we move the "private" methods to the bottom of the file?

nextmv/nextmv/local/executor.py
```diff
@@ -190,6 +191,35 @@ def execute_run(
...
+    if manifest.execution is not None and manifest.execution.entrypoint is not None:
+        return manifest.execution.entrypoint
+
+    # Determine default entrypoint based on type
+    if manifest.type == ManifestType.PYTHON:
+        return "./main.py"
+    elif manifest.type == ManifestType.GO:
+        return "./main"
+    elif manifest.type == ManifestType.JAVA:
+        return "./main.jar"
```
sebastian-quintero: I wonder if this logic actually belongs in the Manifest itself? Probably not.

nextmv/nextmv/manifest.py
```diff
@@ -1005,6 +1005,40 @@ class ManifestConfiguration(BaseModel):
...
 
 
+class ManifestExecution(BaseModel):
+    """
+    Execution configuration for the decision model.
+
+    You can import the `ManifestExecution` class directly from `nextmv`:
+
+    ```python
+    from nextmv import ManifestExecution
```
sebastian-quintero, line 1015: This would be missing in the `__init__.py` file of `nextmv`.

sebastian-quintero (changes requested):
Only blocking because of the missing import in `__init__.py`. The example breaks if this is not there. I would have approved but I fear that the auto-merge is on and thus this gets merged without that change 😅. Sorry for being psycho.

nextmv/nextmv/local/executor.py
merschformann: Sure, I just put it next to the other comparable method. :relaxed:

nextmv/nextmv/manifest.py
merschformann: Good catch. Sorry for missing it. :grimacing:
It worked (for pushing) but of course breaks the docs and the general approach.

merschformann:
Getting failures in unrelated tests @sebastian-quintero . :thinking:

sebastian-quintero:
> Getting failures in unrelated tests @sebastian-quintero . :thinking: 

Weird, looks like it is something wrong with nextmv-gurobipy. I am getting the same failures in my other open PR. I can take a look later in the afternoon, but my best guess is that a new Gurobi release is breaking us, somehow.

sebastian-quintero (approved):
Thanks for the init fix!

PR #162, opened 2025-11-12
Support Gurobi 13 and Nextmv 0.34.1

sebastian-quintero:
# Description

Given that our requirements are set as `>=`, a newer release of `gurobipy` (13.0.0) made our current CI/CD workflows incompatible. With this we add support for newer versions of `gurobipy` and `nextmv`.


sebastian-quintero:
Dropped in favor fo #163

PR #164, opened 2025-11-12
Reduces GH check workload

merschformann:
# Description

Reduces workload of Github checks by only linting on the min version and testing on min and max versions.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.13"]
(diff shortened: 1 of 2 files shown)


.github/workflows/python-test.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.10", "3.13"]
         platform: [ubuntu-latest, windows-latest, macos-latest, macos-15-intel]
```
sebastian-quintero: Do want to also add arch?

sebastian-quintero (approved):
LGTM

.github/workflows/python-test.yml
merschformann: Oh yes, good catch. Sorry, forgot that we also talked about that one. One sec.

PR #174, opened 2025-12-10
Adds `integration_id` to configuration of runs and instances

sebastian-quintero:
# Description

Enables one to run with an `integration_id`. Adds some minor clarification to docs around licensing.

### DIFF
--- a/docs/nextmv/local/index.md
+++ b/docs/nextmv/local/index.md
@@ -3,10 +3,10 @@
 The `nextmv.local` package provides functionality to run Nextmv Applications
 locally on your machine.
 
-!!! success "Free to use"
+!!! success "Open source & free"
 
-    The `local` experience is completely free of charge and does not require a
-    Nextmv account.
+    The `local` experience is completely open-source and free of charge, it does
+    not require a Nextmv account.
 
 At Nextmv, we believe that decision models should be easy to build, test, and
 run. The Nextmv [Cloud Platform][cloud-overview] provides great functionality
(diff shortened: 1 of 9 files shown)


merschformann (approved):
Looks good, thanks for the quick work! 🤗

docs/nextmv/local/index.md
```diff
@@ -3,10 +3,10 @@
 The `nextmv.local` package provides functionality to run Nextmv Applications
 locally on your machine.
 
-!!! success "Free to use"
+!!! success "Open source & free"
```
cowanwalls: Nit: consider updating this to read `"Open source & free of charge"`.

This reads a bit awkwardly as it reads like FOSS but flipped. I'm also not sure if our python package is considered "free software" as far as FOSS nomenclature is concerned as IIRC there's stipulations to that beyond being free of charge, so I'm not sure we could say "Free and open source" verbatim with confidence. I'd need a better understanding of how our licenses fit together to confirm any of that though, which is why I suggest the specific change I did here

```suggestion
!!! success "Open source & free of charge"
```

cowanwalls: If we can say FOSS with confidence (we might, legal stuff and community nomenclature is just something I can't dive into too much detail on reviewing at the moment) we should use the more standard nomenclature, IMO.

```suggestion
!!! success "Free & open source"
```

cowanwalls (approved):
Nitpicks to potentially align on nomenclature. I don't think it's blocking, though.

docs/nextmv/local/index.md
sebastian-quintero: FOSS is accurate, I changed it for Free & open-source

PR #177, opened 2025-12-12
CRUD operations for integrations

sebastian-quintero:
# Description

Adds support for creating, reading, updating and deleting integrations. Provides docs as well.

### DIFF
--- a/docs/nextmv/cloud/integrations.md
+++ b/docs/nextmv/cloud/integrations.md
@@ -77,7 +77,7 @@ integration = cloud.Integration.new(
 print(f"Created integration: {integration.integration_id}")
 ```
 
-the `exist_ok` parameter can be set to `True` to instantiate the integration if
+The `exist_ok` parameter can be set to `True` to instantiate the integration if
 it already exists.
 
 ## Running with an integration
--- a/docs/nextmv/cloud/integrations.md
+++ b/docs/nextmv/cloud/integrations.md
@@ -36,7 +36,7 @@ client = cloud.Client(api_key=os.getenv("NEXTMV_API_KEY"))
 integration = cloud.Integration.new(
     client=client,
     name="My Databricks Runtime",
-    integration_id="my-dbx-runtime",
+    integration_id="my-dbx-runtime", # Auto-generated if omitted
     description="Databricks integration for production workloads",
     integration_type=cloud.IntegrationType.RUNTIME,
     exec_types=[cloud.ManifestType.PYTHON],
@@ -65,6 +65,7 @@ client = cloud.Client(api_key=os.getenv("NEXTMV_API_KEY"))
 integration = cloud.Integration.new(
     client=client,
     name="App-Specific Integration",
+    integration_id="my-dbx-runtime", # Auto-generated if omitted
     integration_type=cloud.IntegrationType.DATA,
     exec_types=[cloud.ManifestType.PYTHON],
     provider=cloud.IntegrationProvider.DBX,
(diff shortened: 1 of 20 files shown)


docs/nextmv/cloud/integrations.md
```diff
@@ -0,0 +1,290 @@
...
+    provider=cloud.IntegrationProvider.DBX,
+    provider_config={"config_key": "config_value"},
+    is_global=False,
+    application_ids=["app-id-1", "app-id-2"],
+)
+
+print(f"Created integration: {integration.integration_id}")
+```
+
+the `exist_ok` parameter can be set to `True` to instantiate the integration if
```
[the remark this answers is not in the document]
cowanwalls: This should be done, yes.

nextmv/nextmv/cloud/integration.py
```diff
@@ -0,0 +1,524 @@
...
+                method="GET",
+                endpoint=f"v1/integrations/{integration_id}",
+            )
+            response_dict = response.json()
+            response_dict["client"] = client
+            integration = cls.from_dict(response_dict)
+
+            return integration
+        except requests.HTTPError as e:
+            raise Exception(f"Failed to retrieve integration '{integration_id}': {e}") from e
```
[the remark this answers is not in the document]
cowanwalls: Assuming this is correct, it's worth considering how we're handling this error case

nextmv/nextmv/cloud/integration.py
```diff
@@ -0,0 +1,524 @@
...
+        ...     provider_config={"config_key": "config_value"},
+        ... )
+        >>> print(integration.to_dict())
+        """
+
+        if integration_id is None:
+            integration_id = safe_id("integration")
+
+        if exist_ok:
+            return cls.get(client=client, integration_id=integration_id)
```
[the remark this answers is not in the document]
cowanwalls: I believe this comment is correct.

nextmv/nextmv/cloud/integration.py
```diff
@@ -0,0 +1,524 @@
...
+
+        if provider_config is not None:
+            payload["provider_config"] = provider_config
+
+        if description is not None:
+            payload["description"] = description
+
+        payload["global"] = is_global
+        if application_ids is not None:
+            payload["application_ids"] = application_ids
```
cowanwalls: I think there's an error case here:

1. Existing integration is already set up for apps
2. We attempt to set it to global (this requires there to be no application IDs on it)
    1. We feed in `true` for the `is_global` value and `None` for the `application_ids` value (expected input)
3. The payload has already had the `application_ids` field modified by the current integration (lines 436-437 above)
4. The code does not override the active `application_ids` in the payload because we passed in `None` (which we expected to do)
5. HTTP request will fail

Proposal:
Update logical statement on line 436 to include a check for `not is_global` before initializing with current `application_ids`

nextmv/nextmv/cloud/integration.py
```diff
@@ -0,0 +1,524 @@
...
+        endpoint="v1/integrations",
+    )
+    response_dict = response.json()
+    integrations = []
+    for integration_data in response_dict.get("items", []):
+        integration_data["client"] = client
+        integration = Integration.from_dict(integration_data)
+        integrations.append(integration)
+
+    return integrations
```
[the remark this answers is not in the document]
cowanwalls: Not pressing, but testing is nice to have where we can.

cowanwalls (changes requested):
A few things caught by Copilot should be looked into or updated. There was another potential error case not caught by Copilot that should also be addressed. Tests would be nice but are not strictly necessary to get this in.

nextmv/nextmv/cloud/integration.py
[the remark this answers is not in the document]
sebastian-quintero: Yes, will add integration tests further down the line.

nextmv/nextmv/cloud/integration.py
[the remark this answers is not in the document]
sebastian-quintero: Thank you helpful robot.

nextmv/nextmv/cloud/integration.py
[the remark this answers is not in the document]
sebastian-quintero: You are correct helpful robot.

docs/nextmv/cloud/integrations.md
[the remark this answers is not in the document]
sebastian-quintero: Thank you little bot.

nextmv/nextmv/cloud/integration.py
sebastian-quintero: I fixed the logic to add a validation step before submitting the request.

nextmv/nextmv/cloud/integration.py
```diff
@@ -0,0 +1,545 @@
...
+        payload = {
+            "name": integration_dict["name"],
+            "id": integration_dict["id"],
+            "description": integration_dict["description"],
+            "global": integration_dict["global"],
+            "type": integration_dict["type"],
+            "provider": integration_dict["provider"],
+            "exec_types": integration_dict["exec_types"],
+            "provider_config": integration_dict["provider_config"],
+        }
```
merschformann: Why are we filtering out the values here? If we would simply keep the original ones, we would not eventually delete user specified values when new fields get introduced.
I.e., `integration_dict` is already the payload, no?

I think we had this discussion before. Maybe I am missing a point here / am forgetting something. 🤔

merschformann (approved):
I have a question about how we do updates, but otherwise all fine. And I don't need to be blocking about it.

Thanks for the work! 🤗

nextmv/nextmv/cloud/integration.py
sebastian-quintero: I don't understand the question...? This is a `PUT` operation, so we are interfacing with it to turn it into a `PATCH` operation. That means that we get all the values the way they are _right now_, and only update those that the user wants to.

merschformann: Exactly, but the marked lines explicitly grab certain fields. When non-breaking changes are added later, these new fields won't get copied. So, why not simply keep the `integration_dict`, modify it and send it back. Why create a new `payload` dict? :thinking:

sebastian-quintero: I understand the question now. I don't know, it was to keep in style with what the other `update` methods are doing. But yeah, I think using the `dict` directly is the better choice.

merschformann: > I don't know, it was to keep in style with what the other update methods are doing.

Yeah, I think we need to update the other ones as well. This came up before, but I think we never got around to it. No worries, just noticing and trying to get us on that more patch-like track. :relaxed:

sebastian-quintero: Ok @merschformann I applied this suggestion here and in the other `update` methods using `PUT` operations.

merschformann: Thank you! :heart:

cowanwalls (approved):

PR #186, opened 2026-01-07
`nextmv cloud run create` command

sebastian-quintero:
# Description

Introduces the `nextmv cloud run create` command with all its subtleties. Additionally, implements `cloud.Application.run_logs_with_polling` to have real time polling against the `logs/live` endpoint. This is used in the new command.

Moved all the "main" command tree apps to the `__init__.py` files in each package, following guides from Typer.

Adds a new `sleep_duration_func` argument to polling options, which allows one to conditionally manipulate how long to sleep for. It is used when polling for run logs, we sleep the amount of seconds dictated by the Cloud API through the `next_available_in_seconds` field.


nextmv/nextmv/cli/configuration/config.py
```diff
@@ -89,15 +90,15 @@ def build_client(profile: str | None = None) -> Client:
 
     if profile is not None:
         if profile not in config:
-            error(f"Profile [bold magenta]{profile}[/bold magenta] does not exist.")
+            error(f"Profile [magenta]{profile}[/magenta] does not exist.")
 
         api_key = config[profile].get(API_KEY_KEY)
         if api_key is None or api_key == "":
-            error(f"API key for profile [bold magenta]{profile}[/bold magenta] is not set or is empty.")
+            error(f"API key for profile [magenta]{profile}[/magenta] is not set or is empty.")
```
merschformann, line 97: We might need a helper or some guideline on styling messages. Not sure. Just noticing. I like the new more colorful output though. 😊

merschformann: For ourselves I mean of course. ☺️

nextmv/nextmv/cli/configuration/config.py
```diff
@@ -110,6 +111,40 @@ def build_client(profile: str | None = None) -> Client:
     return Client(api_key=api_key, url=f"https://{endpoint}")
 
 
+def build_app(app_id: str, profile: str | None = None) -> Application:
```
merschformann, line 114: Is this the right place (as in file) for this func? Or, is it just a bit oddly named? 🤔 
Something feels a bit off, but I might just misunderstand it. 😊

nextmv/nextmv/cli/options.py
```diff
@@ -22,3 +22,17 @@
         metavar="PROFILE_NAME",
     ),
 ]
+
+# app_id option - can be used in any command that requires an application ID.
+# Define it as follows in commands or callbacks, as necessary:
+# app_id: AppIDOption
```
merschformann, line 28: ❤️

nextmv/nextmv/polling.py
```diff
@@ -255,23 +263,27 @@ def poll(  # noqa: C901
...
+                delay = polling_options.delay  # Base
+                delay += polling_options.backoff * (2**ix)  # Add exponential backoff.
+                delay += random.uniform(0, polling_options.jitter)  # Add jitter.
+
+            # We cannot exceed the max delay.
+            if delay >= polling_options.max_delay:
+                max_reached = True
+                delay = polling_options.max_delay
+
+            sleep_duration = delay
```
merschformann, line 287:
```suggestion
            # Calculate delay using exponential backoff with jitter.
            if max_reached:
                # If we already reached the maximum, we don't want to further calculate the
                # delay to avoid overflows.
                delay = polling_options.max_delay
            else:
                delay = polling_options.delay  # Base
                delay += polling_options.backoff * (2**ix)  # Add exponential backoff.

            # We cannot exceed the max delay.
            if delay >= polling_options.max_delay:
                max_reached = True
                delay = polling_options.max_delay

            # Add jitter.
            delay += random.uniform(0, polling_options.jitter)

            sleep_duration = delay
```

Just simplifying it a bit. Take it or leave it, not really important of course. Just an opportunity that I noticed. 😊

nextmv/nextmv/cli/cloud/run/create.py
```diff
@@ -0,0 +1,595 @@
...
+        return
+
+    # If logs tailing/fetching is enabled, use it as the polling strategy.
+    rich.print(":hourglass_flowing_sand: Tailing logs and waiting for run to complete...")
+    run_logs = cloud_app.run_logs_with_polling(
+        run_id=run_id,
+        verbose=True,
+        rich_print=True,
+        polling_options=polling_options,
+    )
```
merschformann: This does not support tailing, right? It only prints the logs after the run completes, instead of during the run as they come in. Or am I missing something? 🤔

merschformann: I just now noticed that we print in two places. Here for printing to file and within the func for printing to stdout. It feels a bit complicated, but should be fine, I think. I didn't test all the cases, but I think my concern shouldn't be blocking. ☺️

merschformann (approved):
Thanks for the work. I had some comments, but I don't think anything needs to be blocking. Don't want to slow this one down. 😊 

Thanks for the work! 🤗

nextmv/nextmv/cli/configuration/config.py
sebastian-quintero: Yep! You read my mind, I was planning a `CONTRIBUTING.md` file under `cli` so that people can read the style guide.

nextmv/nextmv/cli/configuration/config.py
sebastian-quintero: It's placed here because it is close to `build_client` (which it depends on) and that one in turn is close to `load_config`. The main idea is to build an app from a profile which is directly related to the `~/.nextmv/config.yaml` file...

nextmv/nextmv/cli/cloud/run/create.py
sebastian-quintero: Can you further elaborate about your main concern here? I didn't get it. The `run_logs_with_polling` gets the logs and allows you to "print in real time" as they come in. As part of the `create` CLI command, those logs _can_ be saved to a file.

nextmv/nextmv/cli/configuration/config.py
sebastian-quintero: I don't know where else to place it though...

nextmv/nextmv/cli/cloud/run/create.py
sebastian-quintero: @merschformann bump.

nextmv/nextmv/cli/configuration/config.py
sebastian-quintero: @merschformann bump.

merschformann: Yeah, no strong opinions and in general fine to later refactor a bit when further structure becomes clear (or, just leave it like it is since you have the better picture of the final thing in your head already :blush: ).
It was just tripping me that the config now seems to be more in focus than the app or the team/account. I just would have expected it somewhere else, but again, not important to address at all. :relaxed:

nextmv/nextmv/cli/cloud/run/create.py
merschformann: All fine, I was just tripping over the structure mainly. This is all fine though. No action here. :relaxed:

PR #191, opened 2026-01-15
Adds versioning to push

sebastian-quintero:
# Description

Changes the default behavior when pushing an app. Whenever you push an app, a new version is created. Added options to skip this behavior and to customize the version attributes if it is created.

### DIFF
--- a/nextmv/nextmv/cli/cloud/app/push.py
+++ b/nextmv/nextmv/cli/cloud/app/push.py
@@ -35,6 +35,46 @@ def push(
             metavar="MANIFEST_PATH",
         ),
     ] = None,
+    no_version: Annotated[
+        bool,
+        typer.Option(
+            "--no-version",
+            "-n",
+            help="The application will be pushed without creating a new version. "
+            "Default is to create a new version on each push.",
+            rich_help_panel="Versioning control",
+        ),
+    ] = False,
+    version_id: Annotated[
+        str | None,
+        typer.Option(
+            "--version-id",
+            "-v",
+            help="Custom version ID when pushing the application. Automatically generated if not provided.",
+            rich_help_panel="Versioning control",
+            metavar="VERSION_ID",
+        ),
+    ] = None,
+    version_name: Annotated[
+        str | None,
+        typer.Option(
+            "--version-name",
+            "-e",
+            help="Custom version name when pushing the application. Automatically generated if not provided.",
+            rich_help_panel="Versioning control",
+            metavar="VERSION_NAME",
+        ),
+    ] = None,
+    version_description: Annotated[
+        str | None,
+        typer.Option(
+            "--version-description",
+            "-r",
+            help="Custom version description when pushing the application. Automatically generated if not provided.",
+            rich_help_panel="Versioning control",
+            metavar="VERSION_DESCRIPTION",
+        ),
+    ] = None,
     profile: ProfileOption = None,
 ) -> None:
     """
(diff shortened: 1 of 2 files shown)


nextmv/nextmv/cloud/application.py
```diff
@@ -2684,6 +2697,47 @@ def push(
...
+        if verbose:
+            if rich_print:
+                rich.print(
+                    f":white_check_mark: Automatically created new version [magenta]{version.id}[/magenta].",
+                    file=sys.stderr,
+                )
+                rich.print_json(data=version_dict)
+            else:
+                log(f'✅ Automatically created new version "{version.id}".')
+                log(json.dumps(version_dict, indent=2))
```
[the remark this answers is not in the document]
sebastian-quintero: thi niw virsiinin bihiviir in thi pish mithid licks tist civirigi. Blah blah blah, mind your business.

nextmv/nextmv/cloud/application.py
```diff
@@ -2534,14 +2534,18 @@ def new_version(
...
         manifest: Manifest | None = None,
         app_dir: str | None = None,
         verbose: bool = False,
         model: Model | None = None,
         model_configuration: ModelConfiguration | None = None,
         rich_print: bool = False,
+        no_version: bool = False,
+        version_id: str | None = None,
+        version_name: str | None = None,
+        version_description: str | None = None,
```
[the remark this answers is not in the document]
sebastian-quintero: Good one, fixed.

nextmv/nextmv/cloud/application.py
```diff
@@ -2684,6 +2708,47 @@ def push(
...
+
+            return
+
+        now = datetime.now(timezone.utc)
+        now_id = now.strftime("%Y%m%d-%H%M%S")
+        now_display = now.strftime("%Y-%m-%dT%H:%M:%SZ")
+        if version_id is None:
+            version_id = safe_id(prefix="version") + f"-{now_id}"
+        if version_name is None:
+            version_name = f"Version pushed at {now_display}"
```
merschformann: This sounds less like a version name but a description instead. Maybe that is preferable. 🤷 
Personally, I would prefer something like `version_id` directly or `f"Version {version_id}"` or `f"Version {now_display}"` if the random slug is annoying. 🤔 
Anyway, not blocking. But we should avoid changing it later. 😊

nextmv/nextmv/cli/cloud/app/push.py
```diff
@@ -35,6 +35,46 @@ def push(
             metavar="MANIFEST_PATH",
         ),
     ] = None,
+    no_version: Annotated[
+        bool,
+        typer.Option(
+            "--no-version",
```
merschformann, line 41: Not for the scope of this PR, but it would be good to have a local configuration that makes this setting permanent. Similar to `git config`, etc. 😊

merschformann (approved):
Thanks for the work. ❤️ 

I am not sure about the default name, but not worth blocking over it.

nextmv/nextmv/cloud/application.py
sebastian-quintero: No problem, agree. I'll change it.

nextmv/nextmv/cli/cloud/app/push.py
sebastian-quintero: Yes, we can configure some default behaviors later that people want to apply when using the CLI generally. I agree, let's punt to later.

PR #194, opened 2026-01-16
Adds the `nextmv cloud secrets` command tree

sebastian-quintero:
# Description

Adds the `nextmv cloud secrets` command tree with the following subcommands:

- `create`
- `delete`
- `get`
- `list`
- `update`

Fixes some logic when creating a secrets collection and not giving it an ID and/or name. Fixes some formatting issues where the enumerations of some classes where not being enclosed in `magenta` for some option help messages.

### DIFF
--- a/nextmv/nextmv/cli/cloud/run/track.py
+++ b/nextmv/nextmv/cli/cloud/run/track.py
@@ -40,7 +40,7 @@ def track(
             "--status",
             "-s",
             help="Status of the tracked run. Allowed values are: "
-            f"{[v.value for v in TrackedRunStatus.__members__.values()]}",
+            f"[magenta]{[v.value for v in TrackedRunStatus.__members__.values()]}[/magenta].",
             metavar="STATUS",
             rich_help_panel="Tracked run configuration",
         ),
--- a/nextmv/nextmv/cli/cloud/run/track.py
+++ b/nextmv/nextmv/cli/cloud/run/track.py
@@ -11,7 +11,7 @@ import typer
 
 from nextmv.cli.cloud.run.create import build_run_config
 from nextmv.cli.configuration.config import build_app
-from nextmv.cli.message import error, in_progress, print_json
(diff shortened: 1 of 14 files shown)


nextmv/nextmv/cli/cloud/run/track.py
```diff
@@ -40,7 +40,7 @@ def track(
             "--status",
             "-s",
             help="Status of the tracked run. Allowed values are: "
-            f"{[v.value for v in TrackedRunStatus.__members__.values()]}",
+            f"[magenta]{[v.value for v in TrackedRunStatus.__members__.values()]}[/magenta].",
```
merschformann: This is probably already in line with style elsewhere, but just wondering whether something like this would be more suitable:

```python
f"[magenta]{', '.join(f'"{v.value}"' for v in TrackedRunStatus)}[/magenta]."
```

I am just wondering whether the brackets seem a bit too technical for docs-like output. I.e., compare this:

```txt
[magenta]['pending', 'running', 'completed', 'failed'][/magenta].
```

vs. this:

```txt
[magenta]"pending", "running", "completed", "failed"[/magenta].
```

Or similar. But just a thought. Totally open to what it is too. 😊

merschformann: Maybe this could also be a common / helper func to keep the style aligned. 🤔 
I am probably overthinking it though. 😅

merschformann (approved):
Nice. Not my deepest review, but it seemed pretty mechanical anyway. Nice touch with the secret input handling.

I had one comment that was just about styling. Feel free to ignore.

Again, thanks for the work! 🤗

nextmv/nextmv/cli/cloud/run/track.py
sebastian-quintero: Really nice suggestion. I implemented a function for getting the values of an enum, applying the `magenta` styling for each value. Way better. Now it is a comma-separated list.

PR #195, opened 2026-01-19
Refactors `cloud/application.py` file and turns it into a directory

sebastian-quintero:
# Description

One of the main classes in the `nextmv` library is `cloud.Application`. The file it was hosted in, `cloud/application.py` was over 4,700 lines long. Using the file often crashed the linter and/or language server, or made them impossibly slow. This PR refactors that class into several subclasses using a mixin strategy. This means that _there should not be any breaking changes_. Import paths are preserved, and logic for methods is not changed. This PR simply aims to split code into multiple, smaller files, while being 100% backwards compatible.


nextmv/nextmv/cloud/application/_acceptance.py
```diff
@@ -0,0 +1,359 @@
...
+from typing import TYPE_CHECKING, Any
+
+import requests
+
+from nextmv.cloud.acceptance_test import AcceptanceTest, Metric
+from nextmv.cloud.batch_experiment import BatchExperimentRun, ExperimentStatus
+from nextmv.polling import DEFAULT_POLLING_OPTIONS, PollingOptions, poll
+
+if TYPE_CHECKING:
+    from . import Application
```
merschformann, line 14: 😬 
tbf, I had to look this up. 😅

merschformann (approved):
Sounds reasonable. I never used this approach before, but seems solid. 🙌 

I didn't do a full review of the code as I assume it's mechnical. I just reviewed the new structure. ☺️

PR #198, opened 2026-01-20
Adds the `nextmv cloud account` command tree

sebastian-quintero:
# Description

Adds the `nextmv cloud account` command tree with the following subcommands:

- `create`
- `delete`
- `get`
- `update`

Adds the `nextmv cloud run queue` command for listing the queued runs in an account.

The `cloud.Account` class received an overhaul with various CRUD operations added to it.


nextmv/nextmv/cli/cloud/run/queue.py
merschformann, line 1: Just for the record, after internal discussions this will be subject to change. ☺️

nextmv/nextmv/cloud/application/__init__.py
```diff
@@ -574,6 +601,48 @@ def push(  # noqa: C901
...
+        ...     print(f"Run ID: {run.id}, Status: {run.status}")
+        Run ID: run-123, Status: queued
+        Run ID: run-456, Status: running
+        Run ID: run-789, Status: queued
+        """
+
+        response = self.client.request(
+            method="GET",
+            endpoint="v1/account/queue",
+        )
```
merschformann: We should probably use the run list endpoint here instead and filter on status instead. But this is an internal change that could even be done later, I think. 😊

merschformann (commented):
Looks good, just waiting on the queue-related modifications. Happy to rubberstamp anytime. ☺️

nextmv/nextmv/cli/cloud/run/queue.py
sebastian-quintero: Implemented changes to the `run list` command so that we don't even have to think about the queue right now.

nextmv/nextmv/cloud/application/__init__.py
sebastian-quintero: Implemented changes to the `run list` command so that we don't even have to think about the queue right now.

merschformann (approved):
Nice! Thank you! ❤️

ryanjoneil (approved):

PR #200, opened 2026-01-21
Adds the `nextmv cloud input-set` command tree

ryanjoneil:
# Description

Adds the `nextmv cloud input-set` command tree with the following subcommands:

- `create`
- `get`
- `list`
- `update`


nextmv/nextmv/cli/options.py
```diff
@@ -106,3 +107,17 @@
...
+SecretsCollectionIDOption = Annotated[
+    str,
+    typer.Option(
+        "--secrets-collection-id",
+        "-s",
+        help="The Nextmv Cloud secrets collection ID to use for this action.",
+        envvar="NEXTMV_SECRETS_COLLECTION_ID",
+        metavar="SECRETS_COLLECTION_ID",
+    ),
+]
```
ryanjoneil, line 192: Was merely alphabetizing.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+    ] = None,
+    limit: Annotated[
+        int | None,
+        typer.Option(
+            "--limit",
+            "-l",
+            help="Maximum number of runs to include (default: 20, max: 20).",
+            metavar="LIMIT",
+        ),
+    ] = None,
```
sebastian-quintero:
```suggestion
    ] = 20,
```

Typer will set it as default 20, so you don't need to specify that in the help.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+            help="End time for filtering runs (RFC3339 format).",
+            metavar="END_TIME",
+        ),
+    ] = None,
+    limit: Annotated[
+        int | None,
+        typer.Option(
+            "--limit",
+            "-l",
+            help="Maximum number of runs to include (default: 20, max: 20).",
```
sebastian-quintero: With the default value defined, we only need to specify the max here.
```suggestion
            help="Maximum number of runs to include (max: [magenta]20[/magenta]).",
```

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+    ] = None,
+    output: Annotated[
+        str | None,
+        typer.Option(
+            "--output",
+            "-o",
+            help="Saves the input set information to this location.",
+            metavar="OUTPUT_PATH",
+        ),
+    ] = None,
```
sebastian-quintero: I haven't been adding the `output` option to `create` operations, only `get`, `update`, and `list` operations. Either remove here or add it everywhere else for consistency. I think we should remove it from here and we can always add it later to all commands if we see it necessary. What do you think?

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+            "-l",
+            help="Maximum number of runs to include (default: 20, max: 20).",
+            metavar="LIMIT",
+        ),
+    ] = None,
+    inputs: Annotated[
+        str | None,
+        typer.Option(
+            "--inputs",
+            help='Inputs for the input set as JSON. Format: \'{"input-1":{"name":"fur", "description":"ball"}}\'',
```
sebastian-quintero: See the `CONTRIBUTING.md` guide for formatting. This makes this fall in line with the `secrets` and `ensemble` commands where help with JSON formatting is provided. The idea is to use the `[magenta] [/magenta]` block when describing variables, values, etc., in command help menus.

```suggestion
            help="Inputs for the input set . Data should be valid [magenta]json[/magenta]. Object format: [magenta]{'input_id': {'name': 'name', 'description': 'description'}}[/magenta].",
```

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+        # to [{"id": "...", "name": "...", "description": "..."}] format
+        inputs_dict = json.loads(inputs)
+        payload["inputs"] = [{"id": k, **v} for k, v in inputs_dict.items()]
+
+    response = client.request(
+        method="POST",
+        endpoint=f"/v1/applications/{app_id}/experiments/inputsets",
+        payload=payload,
+    )
+    input_set_data = response.json()
```
sebastian-quintero: Nice! But we don't need to implement things two times. Please look at other commands for examples. The idea is to build a `cloud.Application` class. All application-related methods are already there, and if they are not, then we need to add them. That means that all of this needs to be changed for something like this:

```python
cloud_app = build_app(app_id=app_id, profile=profile)
input_set = cloud_app.new_input_set(
    id=input_set_id,
    name=name,
    ....
)
```

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+            --instance-id my-instance --start-time "2024-01-01T00:00:00Z" \\
+            --end-time "2024-01-31T23:59:59Z"[/green]
+    """
+
+    client = build_client(profile)
+    in_progress(msg="Creating input set...")
+
+    # Generate a random ID if not provided
+    if input_set_id is None:
+        input_set_id = safe_id("input-set")
```
sebastian-quintero, line 147: See my other comment about not implementing logic here, but merely calling methods from the SDK. This change would need to be added to the `new_input_set` function.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+        str | None,
+        typer.Option(
+            "--instance-id",
+            "-i",
+            help="Instance ID to filter runs from.",
+            metavar="INSTANCE_ID",
+        ),
+    ] = None,
+    run_ids: Annotated[
+        str | None,
```
sebastian-quintero, line 53: Please see other methods like `run/create.py` or `secrets/create.py` where a `list[str]` is defined. The idea with Typer is to allow the user to declare arrays by either:
1. Using the option multiple times. In this case: `--run-ids foo --run-ids bar --run-ids baz`. This is supported natively in Typer.
2. Separating with comma to be a bit backwards compatible.
3. A combination of both.

This means that we would process this `run_ids` variable to construct a final `run_ids_list`. As mentioned, look at the other files for examples of doing this.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+        typer.Option(
+            "--run-ids",
```
sebastian-quintero, line 64:
```suggestion
            "--run-ids",
            "-r",
```

I think we can use this shorthand here.

nextmv/nextmv/cli/cloud/input_set/get.py
```diff
@@ -0,0 +1,66 @@
...
+    its name, description, and the list of inputs it contains.
+
+    [bold][underline]Examples[/underline][/bold]
+
+    - Get an input set.
+        $ [green]nextmv cloud input-set get --app-id my-app --input-set-id my-input-set[/green]
+
+    - Get an input set and save to a file.
+        $ [green]nextmv cloud input-set get --app-id my-app --input-set-id my-input-set \\
+            --output input-set.json[/green]
```
sebastian-quintero: Would be good to continue with the `hare-app` for the app ID for consistency with the other commands.

nextmv/nextmv/cli/cloud/input_set/get.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+
+    client = build_client(profile)
+    in_progress(msg="Getting input set...")
+
+    response = client.request(
+        method="GET",
+        endpoint=f"/v1/applications/{app_id}/experiments/inputsets/{input_set_id}",
+    )
+    input_set_data = response.json()
```
sebastian-quintero: Same comment here about using `cloud.Application.input_set` as opposed to implementing the logic yourself.

nextmv/nextmv/cli/cloud/input_set/list.py
```diff
@@ -0,0 +1,67 @@
...
+    [bold][underline]Examples[/underline][/bold]
+
+    - List all input sets for an application.
+        $ [green]nextmv cloud input-set list --app-id my-app[/green]
+
+    - List all input sets and save to a file.
+        $ [green]nextmv cloud input-set list --app-id my-app --output input-sets.json[/green]
+
+    - List input sets using a specific profile.
+        $ [green]nextmv cloud input-set list --app-id my-app --profile my-profile[/green]
```
sebastian-quintero: Same `hare-app` comment.

nextmv/nextmv/cli/cloud/input_set/list.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    client = build_client(profile)
+    in_progress(msg="Listing input sets...")
+
+    response = client.request(
+        method="GET",
+        endpoint=f"/v1/applications/{app_id}/experiments/inputsets",
+    )
+    input_sets_data = response.json()
```
sebastian-quintero: Same comment about using the `cloud.Application.list_input_sets` method instead of writing the logic yourself.

nextmv/nextmv/cli/cloud/input_set/update.py
```diff
@@ -0,0 +1,126 @@
...
+        # to [{"id": "...", "name": "...", "description": "..."}] format
+        inputs_dict = json.loads(inputs)
+        payload["inputs"] = [{"id": k, **v} for k, v in inputs_dict.items()]
+
+    response = client.request(
+        method="PUT",
+        endpoint=f"/v1/applications/{app_id}/experiments/inputsets/{input_set_id}",
+        payload=payload,
+    )
+    input_set_data = response.json()
```
sebastian-quintero: Similar comment. However, the `cloud.Application.update_input_set` method does not exist. Can you please implement it, and then call it here? Thx.

nextmv/nextmv/cli/cloud/__init__.py
```diff
@@ -19,6 +20,7 @@
 app.add_typer(data_app, name="data")
 app.add_typer(instance_app, name="instance")
 app.add_typer(managed_input_app, name="managed-input")
+app.add_typer(input_set_app, name="input-set")
```
sebastian-quintero:
```suggestion
app.add_typer(data_app, name="data")
app.add_typer(input_set_app, name="input-set")
app.add_typer(instance_app, name="instance")
app.add_typer(managed_input_app, name="managed-input")
```

Alphabetical ordering of commands.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,195 @@
...
+            metavar="NAME",
+        ),
+    ],
+    input_set_id: Annotated[
+        str | None,
+        typer.Option(
+            "--input-set-id",
+            "-s",
+            help="An optional ID for the input set. If not provided, a random ID will be generated.",
+            metavar="INPUT_SET_ID",
```
sebastian-quintero, line 40: Add this for consistency with the other operations when using this option.

```python
envvar="NEXTMV_INPUT_SET_ID",
```

sebastian-quintero (changes requested):
Mostly blocking due to logic being implemented in the CLI layer as opposed to the SDK layer. Great work! Thank you so much.

nextmv/nextmv/cli/cloud/input_set/create.py
ryanjoneil: Note that `mypy` (which is not the default type checker, but still) says `None` is invalid for a `list[str]`. @sebastian-quintero should I go through and change these from this...

```python
    foo: Annotated[
        list[str],
        typer.Option(
            # bar baz
        ),
    ] = None,
```

...to this?

```python
    foo: Annotated[
        list[str],
        typer.Option(
            # bar baz
        ),
    ] = [],
```

ryanjoneil: Never mind, that doesn't seem to work either.

ryanjoneil: This is a type error, because the types don't match:

```python
    foo: Annotated[  
        list[str],  
        typer.Option(  
            # bar baz  
        ),  
    ] = None,  
```

But this is a linter error, because we shouldn't use mutable data for defaults:

```python
    foo: Annotated[  
        list[str],  
        typer.Option(  
            # bar baz  
        ),  
    ] = [],
```

It _seems_ the right thing to do is to use `list[str] | None` as the first parameter, which is silly for other reasons.

ryanjoneil: I'm going to leave this as follows for now, since that's what we have elsewhere. We can change it if we feel inclined.

```python
    foo: Annotated[  
        list[str],  
        typer.Option(  
            # bar baz  
        ),  
    ] = None,  
```

sebastian-quintero: I was checking the [Typer docs](https://typer.tiangolo.com/tutorial/multiple-values/multiple-options/) and doing your first option is the actual recommendation. I just went through the code and I realized that when a `list` is optional (with default `= None`), then the type should be `list[str] | None` instead of _just_ `list[str]`. I changed it for `instance` and `run` in my current open PR #203. Good find!

sebastian-quintero: Maybe we should add `mypy` to the `ruff` checker.

ryanjoneil (commented):
> I haven't been adding the `output` option to `create` operations, only `get`, `update`, and `list` operations. Either remove here or add it everywhere else for consistency. I think we should remove it from here and we can always add it later to all commands if we see it necessary. What do you think? 

I think removing it makes sense.

nextmv/nextmv/cli/cloud/input_set/create.py
ryanjoneil: Removed for now.

nextmv/nextmv/cloud/application/_input_set.py
```diff
@@ -129,7 +129,7 @@ def new_input_set(
         start_time: Optional[datetime]
             Start time of the input set. This is used to filter the runs
             associated with the input set.
-        inputs: Optional[list[ExperimentInput]]
```
ryanjoneil, line 132: Noticed this type was incorrect.

ryanjoneil:
Example commands that work in my account:

#### `nextmv cloud input-set create`
```bash
nextmv cloud input-set create \
    --app-id knapsack-models \
    --name name-$RANDOM \
    --run-ids cplex-dX6hLzmDR \
    --run-ids highs-3vC1YkmvR

nextmv cloud input-set create \
    --app-id knapsack-models \
    --instance-id cplex \
    --name 2026-01-21-$RANDOM-name \
    --end-time "2025-11-13T15:28:11" \
    --start-time "2025-11-13T15:26:00"

nextmv cloud input-set create \
      --input-set-id "past-day-runs-nyc-pizza-$RANDOM" \
      --name "Past Day Runs" \
      --app-id knapsack-models \
      --inputs '[{"id": "input.json-Ug2ULzivg", "name": "input1", "description": "input1 description"}]'
```

#### `nextmv cloud input-set get`

```bash
nextmv cloud input-set get \
    --app-id knapsack-models \
    --input-set-id past-day-runs-nyc-pizza-32427
```

#### `nextmv cloud input-set list`

```bash
nextmv cloud input-set list \
    --app-id knapsack-models
```

#### `nextmv cloud input-set update`

```bash
nextmv cloud input-set update \
    --app-id knapsack-models \
    --input-set-id past-day-runs-nyc-pizza-32427 \
    --name "name $RANDOM" \
    --description "description $RANDOM" \
    --inputs '[{"id": "input.json-Ug2ULzivg", "name": "input1", "description": "input1 description"}, {"id": "input.json-ZAt8LziDg", "name": "input2", "description": "input2 description"}]'
```

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,168 @@
...
+    ] = None,
+    end_time: Annotated[
+        datetime | None,
+        typer.Option(
+            "--end-time",
+            formats=["%Y-%m-%dT%H:%M:%S%z"],
+            help="End time for filtering runs. (example: [magenta]'2024-01-01T00:00:00Z'[/magenta])",
+            metavar="END_TIME",
+        ),
+    ] = None,
```
ryanjoneil, line 87: These are `datetime` flags instead of `str` flags now. Time zone is required. The simplest way to do this is to use RFC3339Z format, which is the same as what the cloud endpoint expects.

nextmv/nextmv/cli/cloud/input_set/create.py
```diff
@@ -0,0 +1,168 @@
...
+        $ [green]nextmv cloud input-set create --app-id hare-app \\
+            --name "Hare Input Set" --run-ids run-1 --run-ids run-2 --run-ids run-3"[/green]
+
+    - Create an input set with a specific ID.
+        $ [green]nextmv cloud input-set create --app-id hare-app --input-set-id hare-input-set \\
+            --name "Hare Input Set" --run-ids run-1 --run-ids run-2 --run-ids run-3"[/green]
+
+    - Create an input set using existing managed inputs.
+        $ [green]nextmv cloud input-set create --app-id hare-app --name "Hare Input Set" \\
+            --inputs '[{"id": "hare-input-1", "name": "hare input", "description": "hare description"}]'[/green]
```
ryanjoneil, line 134: Note I believe this format may be different from what it was in the old CLI. I think this is correct because it matches the endpoint payload format.

sebastian-quintero (approved):
🥳 🥳 🥳 🥳  AMAZING, thank you!

PR #203, opened 2026-01-22
Adds the `nextmv cloud batch` command tree

sebastian-quintero:
# Description

Adds the `nextmv cloud batch` command tree with the following subcommands:

- `create`
- `delete`
- `get`
- `list`
- `metadata`
- `update`

Makes some modifications to formatting in other commands to avoid having long messages in the help of options.  Rectifies the usage of the `--wait` and `--output` flags for acceptance tests, so that they are independent of each other.

### DIFF
--- a/nextmv/nextmv/cli/cloud/acceptance/create.py
+++ b/nextmv/nextmv/cli/cloud/acceptance/create.py
@@ -11,7 +11,7 @@ from nextmv.cli.configuration.config import build_app
 from nextmv.cli.message import enum_values, error, in_progress, print_json, success
 from nextmv.cli.options import AppIDOption, ProfileOption
 from nextmv.cloud.acceptance_test import Comparison, Metric, MetricToleranceType, MetricType, StatisticType
-from nextmv.polling import DEFAULT_POLLING_OPTIONS
+from nextmv.polling import default_polling_options
 
 # Set up subcommand application.
 app = typer.Typer()
@@ -32,7 +32,8 @@ app = typer.Typer()
 
     Use the [code]--wait[/code] flag to wait for the acceptance test to
     complete, polling for results. Use the [code]--output[/code] flag to
-    specify a destination file for the results.
+    specify a destination file for the output obtained. Please note that using
+    the [code]--output[/code] flag will not activate waiting by itself.
 
     [bold][underline]Metrics[/underline][/bold]
 
@@ -243,7 +244,7 @@ def create(
         typer.Option(
             "--output",
             "-o",
-            help="Save the acceptance test results to this location.",
+            help="Save the acceptance test output to this location.",
             metavar="OUTPUT_PATH",
             rich_help_panel="Output control",
         ),
@@ -274,7 +275,7 @@ def create(
     metrics_list = build_metrics(metrics)
 
     # Build the polling options.
-    polling_options = DEFAULT_POLLING_OPTIONS
+    polling_options = default_polling_options()
     polling_options.max_duration = timeout
 
     # Create the acceptance test
--- a/nextmv/nextmv/cli/cloud/acceptance/create.py
+++ b/nextmv/nextmv/cli/cloud/acceptance/create.py
@@ -31,9 +31,8 @@ app = typer.Typer()
     experiment.
 
     Use the [code]--wait[/code] flag to wait for the acceptance test to
-    complete, polling for results. Using the [code]--output[/code] flag will
-    also activate waiting, and allows you to specify a destination file for the
-    results.
+    complete, polling for results. Use the [code]--output[/code] flag to
+    specify a destination file for the results.
 
     [bold][underline]Metrics[/underline][/bold]
 
@@ -244,7 +243,7 @@ def create(
         typer.Option(
             "--output",
             "-o",
-            help="Waits for the acceptance test to complete and saves the results to this location.",
+            help="Save the acceptance test results to this location.",
             metavar="OUTPUT_PATH",
             rich_help_panel="Output control",
         ),
@@ -278,11 +277,8 @@ def create(
     polling_options = DEFAULT_POLLING_OPTIONS
     polling_options.max_duration = timeout
 
-    # Determine if we should wait
-    should_wait = wait or (output is not None and output != "")
-
     # Create the acceptance test
-    if should_wait:
+    if wait:
         in_progress(msg="Creating acceptance test and waiting for results...")
         acceptance_test = cloud_app.new_acceptance_test_with_result(
             candidate_instance_id=candidate_instance_id,
--- a/nextmv/nextmv/cli/cloud/batch/create.py
+++ b/nextmv/nextmv/cli/cloud/batch/create.py
@@ -1,0 +1,454 @@
+"""
+This module defines the cloud batch create command for the Nextmv CLI.
+"""
+
+import json
+from typing import Annotated
+
+import typer
+
+from nextmv.cli.configuration.config import build_app
+from nextmv.cli.message import error, in_progress, print_json, success
+from nextmv.cli.options import AppIDOption, ProfileOption
+from nextmv.cloud.batch_experiment import BatchExperimentRun
+from nextmv.polling import DEFAULT_POLLING_OPTIONS
+
+# Set up subcommand application.
+app = typer.Typer()
+
+
+@app.command()
+def create(
+    app_id: AppIDOption,
+    batch_id: Annotated[
+        str | None,
+        typer.Option(
+            "--batch-id",
+            "-b",
+            help="ID for the batch experiment. Will be generated if not provided.",
+            envvar="NEXTMV_BATCH_ID",
+            metavar="BATCH_ID",
+            rich_help_panel="Batch experiment configuration",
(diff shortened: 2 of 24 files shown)


nextmv/nextmv/cli/cloud/ensemble/create.py
sebastian-quintero, line 1: The help message was moved to an argument of the decorator to use f-strings. The main change here is that we are avoiding long-winded help messages for the options.

nextmv/nextmv/cli/cloud/acceptance/create.py
```diff
@@ -244,7 +243,7 @@ def create(
         typer.Option(
             "--output",
             "-o",
-            help="Waits for the acceptance test to complete and saves the results to this location.",
+            help="Save the acceptance test results to this location.",
```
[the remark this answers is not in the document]
sebastian-quintero: Clarified docs.

nextmv/nextmv/cli/cloud/batch/create.py
```diff
@@ -0,0 +1,454 @@
...
+    output: Annotated[
+        str | None,
+        typer.Option(
+            "--output",
+            "-o",
+            help="Save the batch experiment results to this location.",
+            metavar="OUTPUT_PATH",
+            rich_help_panel="Output control",
+        ),
+    ] = None,
```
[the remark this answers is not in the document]
sebastian-quintero: Clarified the help

nextmv/nextmv/cli/cloud/batch/create.py
```diff
@@ -0,0 +1,454 @@
...
+    Returns
+    -------
+    list[BatchExperimentRun]
+        List of processed runs.
+    """
+    processed_runs = []
+    for ix, item in enumerate(run_data):
+        _validate_run_fields(item, run_str, ix)
+        run = BatchExperimentRun(**item)
+        processed_runs.append(run)
```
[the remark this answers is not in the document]
sebastian-quintero: It is checked above

nextmv/nextmv/cli/cloud/batch/create.py
```diff
@@ -0,0 +1,454 @@
...
+
+    # Build the runs list from the CLI options
+    runs_list = build_runs(runs)
+
+    # Build the option sets from the CLI options
+    option_sets_dict = build_option_sets(option_sets)
+
+    # Build the polling options.
+    polling_options = DEFAULT_POLLING_OPTIONS
+    polling_options.max_duration = timeout
```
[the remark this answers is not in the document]
sebastian-quintero: Good one

nextmv/nextmv/cli/cloud/acceptance/create.py
```diff
@@ -275,14 +275,11 @@ def create(
     metrics_list = build_metrics(metrics)
 
     # Build the polling options.
-    polling_options = DEFAULT_POLLING_OPTIONS
+    polling_options = default_polling_options()
     polling_options.max_duration = timeout
 
-    # Determine if we should wait
-    should_wait = wait or (output is not None and output != "")
```
merschformann, line 282: Shouldn't `--output` imply waiting though? 🤔 I wouldn't expect anyone to be interested in creating an acceptance test and only writing out the response without the results. Also, it can theoretically (probably not practically though) return completed results or incomplete ones depending on the swiftness of the test.
Then again, I don't have a strong opinion.

merschformann (approved):
Looks straightforward. Just a quick thought on the behavior. ☺️ 

Thanks for the work! 🤗

nextmv/nextmv/cli/cloud/acceptance/create.py
sebastian-quintero: This is to fall in place with behavior in other commands, like `run create`, where output simply implies that you want to save the response to a file. In this case, you can use it without `tail` and it will start the test or experiment and return the metadata, which is what the API returns. If you use with `wait`, then it will poll for results and return the completed results. We are basically falling in line with `wait/output` and `tail/logs` options when creating a run, making the options `wait` and `output` independent of each other, in the same way `tail` and `logs` are independent of each other.

merschformann: Yeah, I understood that. I am not sure how useful it is to persist the "output" if it's not done though. I also wonder what will be written out for a `multi-file` when `--wait` is not present. :thinking: 
Am I missing something else? What is written to the log file on `--logs` if it doesn't cause the wait?

sebastian-quintero: 1. If the experiment/test is not done, the output simply shows metadata about the experiment/test, like the ID, created_at, name, etc...
2. For the `cloud run create` command, when `--wait` is not present, the command simply returns the run ID.
3. Looks like if you use the command like `cloud run create -a APP_ID --logs logs.log` it will poll for results, so it _will_ cause waiting. I may have written things down here incorrectly, but in the `run` command help everything should be correctly documented.

merschformann: I don't think that 2 is correct. I just tested it from this branch even and it blocks and waits for the output. It also makes sense to me. What am I missing? :thinking: 
It would have been surprising behavior when used on `multi-file` in particular otherwise.
[This](https://github.com/nextmv-io/nextmv-py/blob/acc60e1fb977fe6d8fe2818fb81e9ab83fb30e43/nextmv/nextmv/cli/cloud/run/get.py#L149-L152) seems to do exactly this, it not only respects wait but also output.

sebastian-quintero: 2 should be correct.
```bash
$ nextmv cloud run create -a 01-22-mf -i inputs
{
  "run_id": "latest-k6QDaEIDR"
}
```

I am running a multi-file app and if I don't use wait, it simply returns the run ID...

merschformann: You did not use `--output` on the call.

sebastian-quintero: I think something is off. I need to investigate furhter.

sebastian-quintero: Ok, somehow I introduced a bug even though I believed I tested everything thoroughly. Fixing in the upcoming PR. Here is the final behavior for run creation:
1. If no flags set: returns the run ID.
2. If `--wait` is set: polls for results until the run finishes. For json, text prints results to stdout. For multi-file, prints metadata to stdout and results are saved to a dir named after the run_id.
3. If `--output` is set: polls for results until the run finishes. For json, text prints results to the specified file. For multi-file, saves results to the specified dir.
4. If both `--wait` and `--output` is set: it is basically the behavior of `--output`.
5. If `--tail` is set: polls for live logs, which is different than polling for a run to complete. The endpoint for live logs is typically slower than querying the metadata endpoint for runs, as the API almost always suggests sleeping for 30 seconds. Prints logs to stderr. Nothing is done with the results.
6. If `--logs` is set: polls for the run to complete, which is faster than waiting for live logs. After it is done saves the logs to the specified file. Nothing is done with the results.
7. If both `--wait` and `--logs` are used: combines the behavior. Logs are tailed and printed to stderr. After the run is done, logs are saved to the specified file.

sebastian-quintero: This behavior, which is what is documented, should be true after the fix:
```
    Use the [code]--wait[/code] flag to wait for the run to complete, polling
    for results. Using the [code]--output[/code] flag will also activate
    waiting, and allows you to specify a destination (file or dir) for the
    output, depending on the content type.

    Use the [code]--tail[/code] flag to stream logs to
    [magenta]stderr[/magenta] until the run completes. Using the
    [code]--logs[/code] flag will also activate waiting, and allows you to
    specify a file to write the logs to.
```

merschformann: Yeah, so `--output` implies waiting. That means that the behavior is different for acceptance with the change I highlighted above, no? :thinking: 

I would still prefer to collapse these commands to just `--output` and `--logs` like discussed briefly before. I don't find the up to 30s faster helpful, I would rather like to have live logs in the file too (enabling listeners on the file to update as content changes - like when you open the file up in vs-code while the run is still going). 

But I don't want to push too much for this. I just wanted to note the inconsistency here. :blush:

sebastian-quintero: In any case, I think you are right and I will probbaly align behavior for batch, acceptance, scenario. If `--output` is specified, waiting is enabled. If nothing is specified, only the ID is returned. As always, thank you for your reviews and remarks, they make me think deeply about these issues. You know, when you are in the middle of it, it is easy to lose sight of logic because it is a but overwhelming to implement all of this from scratch. So, having another set of eyes is super helpful. 🌮 to you.

sebastian-quintero: @merschformann I will go ahead and merge and implement the behavior changes in the next PR.

PR #208, opened 2026-01-26
Removes PATH check for old CLI path

sebastian-quintero:
# Description

Removes the logic that was checking if the old Go CLI install path was present in the `PATH`. Given that this was about cleanup, and didn't actually affect functionality of the Python CLI, then the check is removed altogether.

This got me thinking about prompts in general in a CI/CD environment, where people can easily not use `--yes` or similar flags, or the check that prompts a user to delete the old CLI. The confirmation prompts of the project were refactored to use a single `get_confirmation` function. This function detects if the shell is not interactive and also applies a timeout strategy so it simply returns the default `False` if no answer is given.

### DIFF
--- a/nextmv/nextmv/cli/configuration/config.py
+++ b/nextmv/nextmv/cli/configuration/config.py
@@ -59,6 +59,18 @@ def save_config(config: dict[str, Any]) -> None:
         yaml.safe_dump(config, file)
 
 
+def non_profile_keys() -> set[str]:
+    """
+    Returns the set of keys that are not profile names in the configuration.
+
+    Returns
+    -------
+    set[str]
+        The set of non-profile keys.
+    """
+    return {API_KEY_KEY, ENDPOINT_KEY}
+
+
 def build_client(profile: str | None = None) -> Client:
     """
     Builds a `cloud.Client` using the API key and endpoint for the given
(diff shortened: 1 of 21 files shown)


nextmv/nextmv/cli/confirm.py
```diff
@@ -0,0 +1,57 @@
...
+
+    try:
+        return result_queue.get(timeout=float(timeout))
+
+    except queue.Empty:
+        print("\n", file=sys.stderr)
+        info(
+            msg=f"No response received within [magenta]{timeout}s[/magenta], assuming no ([magenta]n[/magenta]).",
+            emoji=":bulb:",
+        )
```
merschformann: I would personally not do this. It's typical behavior (imho) that CLI tools wait indefinitely when asking for user confirmation. I am not aware of an example using a timeout, but I do know many that just block indefinitely. I think this may be unexpected behavior to users. And it happens easily that one starts a command and context switches to something else only to come back later and running into an unwanted timeout, I think.

I appreciate the idea though and we can discuss this quickly in a small circle. Happy to get overruled, might be just my personal experience being off. 😊

nextmv/nextmv/cli/configuration/config.py
merschformann, line 1: Some nice helpful new hints in this file. ❤️

merschformann (commented):
I have one concern about the new timeout for confirmation behavior. Otherwise, happy to approve. Thanks for the work! 🤗

nextmv/nextmv/cli/confirm.py
sebastian-quintero: Ok, I thought it was a nice user experience to get a timeout, but I will get rid of the timeout 👍🏻 . No problem.

merschformann: Yeah, I like the intention, but to me personally it might be a negative in the end. However, feel free to check with the others. As I said, it might just be me. Not a particularly strong opinion. :relaxed:

merschformann (approved):
❤️

PR #210, opened 2026-01-26
Normalizes clone directory path

sebastian-quintero:
# Description

Normalizes the path when using the `nextmv community clone` command with the `--directory` option so using it like `--directory foo/` is normalized as `--directory foo`.

### DIFF
--- a/nextmv/nextmv/cli/community/clone.py
+++ b/nextmv/nextmv/cli/community/clone.py
@@ -97,8 +97,10 @@ def clone(
     if version == LATEST_VERSION:
         version = app_obj.get("latest_app_version")
 
-    destination = directory
-    if directory is None or directory == "":
+    # Clean and normalize directory path in an OS-independent way
+    if directory is not None and directory != "":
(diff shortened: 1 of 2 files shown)


nextmv/nextmv/cli/community/clone.py
```diff
@@ -97,8 +97,10 @@ def clone(
     if version == LATEST_VERSION:
         version = app_obj.get("latest_app_version")
 
-    destination = directory
-    if directory is None or directory == "":
+    # Clean and normalize directory path in an OS-independent way
+    if directory is not None and directory != "":
```
ryanjoneil, line 101: Would `if directory` work here or is that considered bad form?

ryanjoneil (approved):

nextmv/nextmv/cli/community/clone.py
sebastian-quintero: I like to be explicit with values and types and not rely on boolean evaluation of types. I guess that is typed behavior sticking in me.

PR #214, opened 2026-01-27
Updates help descriptions for `account` and `data`

sebastian-quintero:
# Description

Uses the SSO language more prominently in `account` commands and clarifies that data upload is not used only for runs but for application components in general.


merschformann (approved):
🎉

PR #216, opened 2026-01-28
Adds support for new output field on metrics, deprecates statistics

racquesta:
# Description

Adds support for the new metrics field on output while deprecating statistics.

## CLI
- Adds `--metrics` option to `nextmv cloud run track` command.

## SDK
- Adds metrics field to output.
- Write metrics file to local output if metrics is given in output or for `multi-file` content format.
- Include support for metrics file to be included in `sync` operation when syncing local runs to cloud.


nextmv/nextmv/cli/cloud/acceptance/create.py
```diff
@@ -71,8 +71,8 @@
     [bold][underline]Examples[/underline][/bold]
 
     - Create an acceptance test with a single metric.
-        $ [dim]METRIC='{{
-            "field": "solution.objective",
+        $ [green]METRIC='{{
```
sebastian-quintero: This should be `[dim]` according to the style guide and to stay consistent with the rest of the commands. Here and elsewhere.

nextmv/nextmv/cli/cloud/run/track.py
```diff
@@ -124,11 +124,21 @@ def track(
     statistics: Annotated[
         str | None,
         typer.Option(
-            help="The statistics of the run being tracked. A [magenta]json[/magenta] file to read the statistics from.",
+            help="Deprecated: Use 'metrics' instead. The statistics of the run" \
+                " being tracked. A [magenta]json[/magenta] file to read the statistics from.",
```
sebastian-quintero:
```suggestion
            help="[red](deprecated) Use --metrics instead[/red]. The statistics of the run" \
                " being tracked. A [magenta]json[/magenta] file to read the statistics from.",
```

Typer does not have a built-in argument for marking an option as deprecated, but to try to match the style of command deprecation, let's use `[red]` with the lowercase and parenthesis. Also, when referring to options, just mention the option, and Typer should auto-format for us.

nextmv/nextmv/cli/cloud/run/track.py
sebastian-quintero, line 234: At the top of the command implementation, we should do something similar to this:

```python
if statistics:
    warning("The --statistics option is deprecated, use --metrics instead.")
```

`warning` can be imported from `nextmv.cli.message`.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -1015,6 +1022,18 @@ def track_run(
...
+        metrics = tracked_run.metrics
+        if metrics is not None:
+            if isinstance(metrics, dict):
+                metrics_dict = metrics
+            else:
+                raise ValueError("tracked_run.metrics must be a `dict` object")
+
+            url_metrics = self.upload_url()
+            self.upload_data(data=metrics_dict, upload_url=url_metrics)
+            external_result.metrics_upload_id = url_metrics.upload_id
```
sebastian-quintero: I changed this in #225, you will probably have to resolve conflicts here. The idea is that I moved this to an auxiliary function, along with the other logic for handling input, output, assets.

nextmv/nextmv/default_app/main.py
```diff
@@ -26,12 +26,10 @@
...
     solution={"message": message},
-    statistics=nextmv.Statistics(
-        result=nextmv.ResultStatistics(
-            value=1.23,
-            custom={"message": message},
-        ),
-    ),
+    metrics={
+        "value": 1.23,
+        "custom": {"message": message},
```
sebastian-quintero:
```suggestion
        "message": message,
```

I think the simpler the better.

nextmv/nextmv/local/executor.py
```diff
@@ -508,6 +584,9 @@ def process_run_statistics(
     manifest: Manifest,
 ) -> None:
     """
+    !!! warning
+    `process_run_statistics` is deprecated, use `process_run_metrics` instead.
```
sebastian-quintero: This function is not used by anyone as part of the API. I think we should maybe just adapt to have a single `process_run_metrics` that is able to handle both metrics and statistics?

nextmv/nextmv/output.py
```diff
@@ -89,7 +98,7 @@
 
 class RunStatistics(BaseModel):
     """
-    Statistics about a general run.
+    Deprecated: Statistics about a general run.
```
sebastian-quintero: You should use the `!!! warning` style here, because in the docs it gets rendered correctly. Look at other examples where this warning admonition (as it is called by material for mkdocs) is used. This should be changed here and elsewhere.

```suggestion
    !!! warning
        `RunStatistics` is deprecated, use `metrics` on the `Output` directly. 
    
    Statistics about a general run.
```

nextmv/nextmv/run.py
```diff
@@ -434,6 +434,9 @@ class RunInfoStatistics(BaseModel):
     """List of statistics indicators."""
 
 
+RunInfoMetrics = RunInfoStatistics
```
sebastian-quintero: What does this mean?

nextmv/tests/test_entrypoint/test_entrypoint.py
```diff
@@ -5,6 +5,7 @@
 import sys
 import unittest
 
+import nextmv.cloud
```
sebastian-quintero: Why is this import needed?

sebastian-quintero (changes requested):
Some general notes about style. Almost there! 🚀

sebastian-quintero:
Another comment: can you please make the title meaningful and fill in the description of the PR? The title is used as release notes, so the idea is that it concisely describes what we are adding/fixing here.

nextmv/nextmv/cloud/application/_run.py
sebastian-quintero: Looks like there are no conflicts, but if you take a look at the rest of the function, you'll find the use of the `self.__handle_tracked_run_xxx` methods.

nextmv/nextmv/default_app/main.py
racquesta: ok, my intention was to show a nested object though

nextmv/nextmv/local/executor.py
racquesta: I think we should leave this as is and tease out the metrics versus stats functionality.  We can still write both locally, but on a tracked run or during "sync" we follow the precedence rules that exist on the backend.

nextmv/nextmv/run.py
racquesta: I'm just creating an alias basically...it is the same

nextmv/tests/test_entrypoint/test_entrypoint.py
racquesta: it's not

nextmv/nextmv/local/executor.py
sebastian-quintero:
```suggestion
    !!! warning
        `process_run_statistics` is deprecated, use `process_run_metrics` instead.
```

Still, indenting is off :)

nextmv/nextmv/output.py
```diff
@@ -89,6 +98,9 @@
 
 class RunStatistics(BaseModel):
     """
+    !!! warning
+    `RunStatistics` is deprecated, use `metrics` on the `Output` directly.
```
sebastian-quintero:
```suggestion
    !!! warning
        `RunStatistics` is deprecated, use `metrics` on the `Output` directly.
```

Indenting here and elsewhere is off. See how to use it [here](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage). The content of the warning needs to be indented by 4 spaces.

nextmv/nextmv/output.py
```diff
@@ -129,6 +141,9 @@ class RunStatistics(BaseModel):
 
 class ResultStatistics(BaseModel):
     """
+    !!! warning
+    `ResultStatistics` is deprecated, use `metrics` on the `Output` directly.
```
sebastian-quintero:
```suggestion
    !!! warning
        `ResultStatistics` is deprecated, use `metrics` on the `Output` directly.
```

nextmv/nextmv/output.py
```diff
@@ -169,6 +184,9 @@ class ResultStatistics(BaseModel):
 
 class DataPoint(BaseModel):
     """
+    !!! warning
+    `DataPoint` is deprecated, use `metrics` on the `Output` directly.
```
sebastian-quintero:
```suggestion
    !!! warning
        `DataPoint` is deprecated, use `metrics` on the `Output` directly.
```

nextmv/nextmv/output.py
```diff
@@ -202,7 +220,8 @@ class DataPoint(BaseModel):
 
 class Series(BaseModel):
     """
-    A series of data points for visualization or analysis.
+
+    Deprecated: A series of data points for visualization or analysis.
```
sebastian-quintero: Needs to follow the same deprecated style warning.

nextmv/nextmv/output.py
```diff
@@ -236,6 +255,9 @@ class Series(BaseModel):
 
 class SeriesData(BaseModel):
     """
+    !!! warning
+    `SeriesData` is deprecated, use `metrics` on the `Output` directly.
```
sebastian-quintero:
```suggestion
    !!! warning
        `SeriesData` is deprecated, use `metrics` on the `Output` directly.
```

nextmv/nextmv/output.py
```diff
@@ -271,6 +293,9 @@ class SeriesData(BaseModel):
 
 class Statistics(BaseModel):
     """
+    !!! warning
+    `Statistics` is deprecated, use `Metrics` instead.
```
sebastian-quintero:
```suggestion
    !!! warning
        `Statistics` is deprecated, use `Metrics` instead.
```

nextmv/nextmv/run.py
sebastian-quintero: I think we should create a new class and deprecate `RunInfoStatistics`. We typically don't do alias to use the new language. The `RunInfoStatistics` has "statistics" written all over it, which of course is fine. To be consistent with what we are trying to do with metrics, I would rather have a new class that has metrics written all over it. The alternative is to keep the alias but change the entire language in the `RunInfoStatistics` class to reflect the wording for metrics.

sebastian-quintero (changes requested):
The warning docstrings require formatting changes. Take a look at the comment regarding the class alias. Looking good otherwise.

sebastian-quintero (approved):
🚀

PR #217, opened 2026-01-28
Removes deprecated functionality in preparation for major release `v1.0.0`

sebastian-quintero:
# Description

Removes all code that was marked as `deprecated` in preparation for the major release of `v1.0.0`. When new functionality is marked as deprecated, we should get rid of it in the next major release.


merschformann (approved):
🪦

PR #223, opened 2026-02-03
Adds support for SSO configuration in SDK and CLI, and deprecate `Account.queue` method

sebastian-quintero:
# Description

- Adds the `cloud/sso.py` file containing the `SSOConfiguration` class able to create, get, delete, update, enable, and disable an SSO configuration. This is to add support for the SDK.
- Adds the `cli/cloud/sso` dir containing the `nextmv cloud sso` command tree to support this in the CLI.
- Marks `Account.queue()` method as deprecated and indicates users to prefer `Application.list_runs` with `status`.

### DIFF
--- a/nextmv/nextmv/cli/cloud/sso/create.py
+++ b/nextmv/nextmv/cli/cloud/sso/create.py
@@ -2,6 +2,7 @@
 This module defines the cloud sso create command for the Nextmv CLI.
 """
 
+import sys
 from typing import Annotated
 
 import typer
@@ -48,7 +49,7 @@ def create(
         typer.Option(
             "--metadata-document",
             "-d",
-            help="The SSO metadata document as a string.",
+            help="The SSO metadata document as a string or a path to a file containing the document.",
             metavar="METADATA_DOCUMENT",
         ),
     ] = None,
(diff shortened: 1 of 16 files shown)


nextmv/nextmv/cli/cloud/sso/create.py
```diff
@@ -0,0 +1,99 @@
...
+
+    - Create and enable SSO configuration immediately.
+        $ [dim]nextmv cloud sso create --metadata-url "https://sso.bunnylogistics.io/metadata" \\
+            --enabled[/dim]
+
+    - Create SSO configuration allowing non-domain users.
+        $ [dim]nextmv cloud sso create --metadata-url "https://idp.hopmail.com/saml/metadata" \\
+            --allow-non-domain-users[/dim]
+
+    - Create SSO configuration using a metadata document string.
```
chriskeyser, line 87: I think the way I've done this is to provide a file that was read for the document. Unclear if this is the case though here? It would work if piped in too, just didn't do it that way. The docs can be somewhat long.

nextmv/nextmv/cli/cloud/sso/delete.py
```diff
@@ -0,0 +1,58 @@
...
+    - Delete the SSO configuration.
+        $ [dim]nextmv cloud sso delete[/dim]
+
+    - Delete the SSO configuration without confirmation prompt.
+        $ [dim]nextmv cloud sso delete --yes[/dim]
+    """
+
+    if not yes:
+        confirm = get_confirmation(
+            "Are you sure you want to delete the sso configuration? This action cannot be undone.",
```
chriskeyser: Could someone infer that if the delete it, they can't reconfigure with the same settings? I think this is a little different than deleting an entity (which deletes dependencies) in that regard.

chriskeyser (approved):
couple of minor comments, none blocking. Nice work!

nextmv/nextmv/cli/cloud/sso/create.py
sebastian-quintero: Good call, it only allowed to pass in the file contents as string to the metadata document option. Not it can be stdin, file string, or file path.

nextmv/nextmv/cli/cloud/sso/delete.py
sebastian-quintero: Yeah true, thanks. I updated the documentation.

PR #232, opened 2026-02-18
Move `all` optional deps to mandatory deps

sebastian-quintero:
# Description

Moves `folium` and `plotly` as mandatory dependencies so that we can run `nextmv local run` on a fresh install of the CLI.


merschformann (approved):
I hope the additional dependencies don't cause too much trouble for the plain modeling / SDK use case. 😬

PR #237, opened 2026-02-19
MCP server for agents

ryanjoneil:
## Summary

image

Adds an MCP (Model Context Protocol) _preview_ server (see above) to the Nextmv Python CLI, enabling any MCP-compatible client (Claude Code, Cursor, VS Code, Copilot, Codex, etc.) to interact with Nextmv Cloud and local applications through natural language.

```
       -------
👦 --> | CLI |
       ------- \
                \    --------------     ----------------
                 --> | Python SDK | --> | Nextmv Cloud |
                /    --------------     ----------------
       ------- /
🤖 --> | MCP |
       -------
```

### What's included

- **`nextmv mcp serve` command** — starts the MCP server with `stdio` (default) or `streamable-http` transport. Registration is as simple as `claude mcp add nextmv -- nextmv mcp serve`.
- **20 tool modules** covering the full Nextmv Cloud surface: apps, runs, versions, instances, batch experiments, input sets, acceptance tests, scenarios, ensembles, shadow tests, switchback tests, secrets, SSO, account info, managed inputs, and profiles — plus local run management, community marketplace browsing, and a workflow guide.
- **Optional dependency** — MCP is gated behind an optional extra (`pip install nextmv[mcp]`). The `mcp serve` subcommand only appears when installed. `mcp[cli]>=1.0` is the sole new dependency.
- **Extensive tests** covering all tool modules, helpers, cloud run caching, profile management, and the serve command.

### Version

Bumps version to `v1.3.1.dev3`.

### Context window management

LLM context windows are finite and expensive. The MCP server is designed around keeping payloads out of the conversation:

- **File-based responses** — large payloads (run results, inputs, logs) are written to local files instead of returned inline. The tool response contains only the file path, so the agent can selectively read what it needs.
- **Cloud run local cache** — cloud run data is cached at `~/.nextmv/runs/{endpoint}/{run_id}/` using the same directory layout as local runs (`inputs/`, `outputs/solutions/`, `logs/`, etc.). Subsequent reads hit the cache and skip the API call entirely. This also means agents can browse cached data with normal file-reading tools without calling MCP tools again.
- **Structured output extraction** — when a cloud run result is downloaded, outputs are split into component files (solutions, metrics, statistics, assets) mirroring the local run layout, so the agent can read just the metrics without loading the full solution.

### LLM input normalization

LLMs are unreliable parameter generators. They frequently pass `""` instead of omitting optional fields, send `"min"` instead of `"minimize"`, or include trailing whitespace. Rather than surfacing cryptic SDK errors, the server normalizes inputs at the boundary:

- `_none_if_empty()` converts empty/whitespace-only strings to `None`.
- `_require_non_empty()` raises clear errors for required fields.
- Enum shorthands (e.g. `"min"`/`"max"` → `"minimize"`/`"maximize"` for ensemble rules, bare floats → `RuleTolerance` dicts) are accepted and expanded before reaching the SDK.

### Profile management

The server supports multiple Nextmv Cloud profiles from `~/.nextmv/config.yaml`. Agents can list profiles, switch the active profile mid-session, and all subsequent tool calls route to the selected profile's endpoint and API key. Profile state is stored in a `contextvars.ContextVar` so concurrent HTTP requests (streamable-http transport) are isolated.

### Workflow guide

A `nextmv_workflow_guide` tool returns a structured guide covering app scaffolding, manifest configuration, coding patterns for JSON and multi-file formats, local testing, Cloud deployment, and experiment setup. This gives agents the context they need to build and deploy apps end-to-end without relying on external documentation.

### Divergence from the CLI

The MCP tools intentionally diverge from the CLI in several areas to better serve agent workflows:

- **Ensembles** — the CLI has no ensemble commands. The MCP server exposes full ensemble CRUD (`cloud_create_ensemble`, `cloud_list_ensembles`, `cloud_get_ensemble`, `cloud_delete_ensemble`) plus dedicated `cloud_ensemble_run` and `cloud_ensemble_run_submit` tools. These handle `RunTypeConfiguration` setup internally so the agent just passes an `ensemble_id` instead of manually wiring run type config. The tolerance and objective fields accept LLM-friendly shorthands.
- **Community apps** — the CLI's `community` commands are mirrored but simplified: `community_list` and `community_clone` with no interactive prompts.
- **Run submit/poll split** — both cloud and local runs offer a `_submit` variant that returns the run ID immediately and a separate poll/result tool, giving agents the ability to fire-and-forget or run multiple jobs concurrently.
- **Managed inputs** — run tools accept a `managed_input_id` parameter directly, which the CLI handles through separate flags.

### Tool registration pattern

Each domain (apps, runs, versions, etc.) lives in its own module under `tools/`. Every module exports a `register(mcp)` function that defines `@mcp.tool()` closures. The top-level `server.py` calls each `register()` in sequence. This keeps the tool count high (~45 tools) without a monolithic file, and makes it easy to add new tool domains.

### Shared helpers

All tool modules import from `tools/_helpers.py`, which provides:
- `ProfileSession` — singleton managing the active profile via `contextvars`.
- `_get_client()` / `_get_app()` — convenience wrappers that respect the active profile.
- `_save_to_file()` / `_save_cloud_run_file()` — serialization to temp or cache files.
- `_extract_cloud_run_outputs()` — splits a cloud result dict into the local run directory structure.
- `_build_run_configuration()` — builds `RunConfiguration` from a content format string.


nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+import os
+import tempfile
+from typing import Any
+
+from nextmv import local
+from nextmv.cloud import Application, Client
+from nextmv.local.local import LOGS_FILE, LOGS_KEY
+from nextmv.output import ASSETS_KEY, METRICS_KEY, OUTPUTS_KEY, SOLUTIONS_KEY, STATISTICS_KEY
+
+logger = logging.getLogger(__name__)
```
ryanjoneil: I think we should use `nextmv.logger.log` instead of the `logging` module.

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+        Args:
+            profile: Optional profile name override. If not provided, uses
+                the session-level profile.
+        """
+
+        resolved = profile or self.profile
+
+        api_key = os.getenv("NEXTMV_API_KEY")
+        if api_key and not resolved:
+            endpoint = os.getenv("NEXTMV_ENDPOINT", "https://api.cloud.nextmv.io")
```
ryanjoneil: Extract the `"https://api.cloud.nextmv.io"` string to a `DEFAULT_NEXTMV_ENDPOINT` variable at the package level.

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+
+
+def _build_run_configuration(content_format: str | None):
+    """Build a RunConfiguration for the given content format string, or None."""
+
+    if content_format is None:
+        return None
+
+    from nextmv.input import InputFormat
+    from nextmv.run import Format, FormatInput, RunConfiguration
```
ryanjoneil: Move imports to the top.

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+        api_key = os.getenv("NEXTMV_API_KEY")
+        if api_key and not resolved:
+            endpoint = os.getenv("NEXTMV_ENDPOINT", "https://api.cloud.nextmv.io")
+            if not endpoint.startswith("http"):
+                endpoint = f"https://{endpoint}"
+            return Client(api_key=api_key, url=endpoint)
+
+        # Fall back to the CLI configuration file (~/.nextmv/config.yaml).
+        try:
+            from nextmv.cli.configuration.config import build_client
```
ryanjoneil: Move imports to the top.

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+session = ProfileSession()
+
+
+def _mask_key(key: str | None) -> str | None:
+    """Mask an API key, keeping only the last 4 characters visible."""
+
+    if not key:
+        return None
+    if len(key) <= 4:
+        return key
```
ryanjoneil: This should mask the entire key.
```suggestion
    if len(key) <= 4:
        return "X" * len(key)
```

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+        return None
+
+    from nextmv.input import InputFormat
+    from nextmv.run import Format, FormatInput, RunConfiguration
+
+    config = RunConfiguration()
+    config.format = Format(
+        format_input=FormatInput(
+            input_type=InputFormat(content_format),
+        ),
```
ryanjoneil:
```suggestion
        format_input=FormatInput(input_type=InputFormat(content_format)),
```

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+
+    This keeps large payloads out of the MCP response (and thus out of
+    the LLM context window). The caller can selectively read the file
+    using standard file-reading tools.
+    """
+
+    fd, path = tempfile.mkstemp(suffix=".json", prefix=f"{prefix}_")
+    with os.fdopen(fd, "w") as fh:
+        json.dump(data, fh, indent=2)
+    return f"Data saved to {path} — use file-reading tools to inspect the contents."
```
ryanjoneil: I think this should be more generic than JSON. The caller should set the suffix, and we shouldn't have to reformat.

If the indentation is necessary, we can update the existing logic to be a `_save_to_json_file` function.

```suggestion
def _save_to_file(data: Any, prefix: str, suffix: str) -> str:
    """Serialize data to a temp JSON file and return a message with the path.

    This keeps large payloads out of the MCP response (and thus out of
    the LLM context window). The caller can selectively read the file
    using standard file-reading tools.
    """

    fd, path = tempfile.mkstemp(prefix=f"{prefix}_", suffix=suffix)
    with os.fdopen(fd, "w") as fh:
        fh.write(data)
    return f"Data saved to {path} — use file-reading tools to inspect the contents."
```

nextmv/nextmv/cli/mcp/tools/_helpers.py
```diff
@@ -0,0 +1,316 @@
...
+    assets = output.get(ASSETS_KEY)
+    if assets:
+        path = os.path.join(outputs_dir, ASSETS_KEY, f"{ASSETS_KEY}.json")
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        with open(path, "w") as fh:
+            json.dump({ASSETS_KEY: assets}, fh, indent=2)
+
+    # Generate visuals from assets (Plotly/GeoJSON → HTML), matching local.
+    try:
+        from nextmv.local.executor import process_run_visuals
```
ryanjoneil: Move imports to top when possible.

nextmv/nextmv/cli/mcp/tools/ensemble.py
```diff
@@ -0,0 +1,344 @@
...
+    """
+
+    from nextmv.cloud.ensemble import (
+        EvaluationRule,
+        RuleObjective,
+        RuleTolerance,
+        RuleToleranceType,
+    )
+
+    # Normalise the tolerance field: the LLM may pass a bare
```
ryanjoneil:
```suggestion
    # Normalize the tolerance field: the LLM may pass a bare
```

nextmv/nextmv/cli/mcp/tools/ensemble.py
```diff
@@ -0,0 +1,344 @@
...
+    app_id: str,
+    ensemble_id: str,
+    content_format: str | None,
+) -> tuple[Application, RunConfiguration]:
+    """Build the Application and RunConfiguration for an ensemble run.
+
+    Raises ``ValueError`` if required parameters are empty.
+    """
+
+    from nextmv.run import RunType, RunTypeConfiguration
```
ryanjoneil: Move imports to the top.

nextmv/nextmv/cli/mcp/tools/ensemble.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+
+        input_dir_path = _helpers._none_if_empty(input_dir_path)
+        managed_input_id = _helpers._none_if_empty(managed_input_id)
+        content_format = _helpers._none_if_empty(content_format)
+
+        try:
+            app, config = _build_ensemble_run_config(app_id, ensemble_id, content_format)
+        except ValueError as e:
+            return str(e)
```
ryanjoneil: Is this the correct thing to do here? What happens if the error isn't caught? If converting it to a string is the correct thing to do, should we add more information?

nextmv/nextmv/cli/mcp/tools/ensemble.py
```diff
@@ -0,0 +1,344 @@
...
```
ryanjoneil: Same comment re: error handling.

nextmv/nextmv/cli/mcp/tools/guide.py
```diff
@@ -0,0 +1,423 @@
...
+- [ ] At least one Plotly visualization as `nextmv.Asset`
+- [ ] Dependencies are Apache 2.0 / MIT / BSD licensed
+- [ ] `input.json` and `inputs/` directory with 3-5 varied inputs
+- [ ] `requirements.txt` with pinned versions
+- [ ] Local runs succeed with expected metrics
+- [ ] At least one non-default option tested
+- [ ] Synced to Cloud, pushed version, created instances
+- [ ] Acceptance test comparing staging vs production reviewed
+- [ ] **User confirmed** before promoting to production
+"""
```
ryanjoneil: If possible, it would be nice for this to live in a `WORKFLOW.md` file and load that in on request rather than embed it directly in the `.py`. That would make it easier to edit.

nextmv/nextmv/cli/mcp/tools/profile.py
```diff
@@ -0,0 +1,81 @@
...
+    """Register profile management tools."""
+
+    @mcp.tool()
+    def cloud_list_profiles() -> list[dict[str, str | None]]:
+        """List available Nextmv Cloud profiles.
+
+        Reads profiles from ``~/.nextmv/config.yaml``. Each returned
+        profile contains a name, endpoint URL, and a masked API key
+        (only the last 4 characters are visible). The ``"default"``
+        profile corresponds to the top-level configuration keys.
```
ryanjoneil: Update this based on the suggested change. Keys 4 characters or less are all Xs.

nextmv/nextmv/cli/mcp/tools/run.py
```diff
@@ -0,0 +1,405 @@
...
+from nextmv.safe import safe_id
+
+
+def _cloud_list_runs_impl(
+    app_id: str,
+    status: str | None = None,
+) -> list[dict[str, Any]]:
+    """Implementation for listing cloud runs."""
+
+    from nextmv.status import StatusV2
```
ryanjoneil: Move imports to the top when possible.

nextmv/nextmv/cli/mcp/tools/sso.py
```diff
@@ -0,0 +1,29 @@
...
+        Removes a domain from the current Nextmv Cloud organization's
+        SSO configuration. After deletion, users with email addresses
+        in that domain will no longer be redirected to the SSO
+        identity provider.
+
+        Args:
+            domain: The domain to remove (e.g., ``"example.com"``).
+        """
+
+        from nextmv.cloud.sso import SSOConfiguration
```
ryanjoneil: Move imports to the top when possible.

nextmv/nextmv/cli/mcp/tools/switchback.py
```diff
@@ -0,0 +1,162 @@
...
+                in minutes (e.g., ``60.0`` for one-hour units).
+            units: Total number of switchback units (alternating
+                periods between baseline and candidate).
+            switchback_test_id: Optional test ID. Auto-generated if
+                omitted.
+            name: Optional human-readable name.
+            description: Optional description.
+        """
+
+        from nextmv.cloud import TestComparisonSingle
```
ryanjoneil: Move imports to the top when possible.

nextmv/nextmv/cli/mcp/tools/switchback.py
```diff
@@ -0,0 +1,162 @@
...
+        promote the candidate instance to production.
+
+        Args:
+            app_id: The application ID.
+            switchback_test_id: The switchback test ID to stop.
+            intent: Stop intent. Allowed values: ``"cancel"``
+                (default) or ``"promote"``.
+        """
+
+        from nextmv.cloud import StopIntent
```
ryanjoneil: Move imports to the top when possible.

nextmv/nextmv/cli/mcp/serve.py
```diff
@@ -0,0 +1,66 @@
...
+        $ [dim]nextmv mcp serve[/dim]
+
+    - Start the MCP server with HTTP transport on port 9090.
+        $ [dim]nextmv mcp serve --transport streamable-http --port 9090[/dim]
+
+    - Register with Claude Code.
+        $ [dim]claude mcp add nextmv -- nextmv mcp serve[/dim]
+    """
+
+    from nextmv.cli.mcp.server import create_server
```
ryanjoneil: Move imports to the top when possible.

nextmv/nextmv/__about__.py
```diff
@@ -1 +1 @@
-__version__ = "v1.3.0"
```
ryanjoneil: If we're OK with it, I'd like to merge this into `develop` and release. If it seems to risky, I can create another `dev` tag for the preview release.

nextmv/nextmv/cloud/batch_experiment.py
```diff
@@ -260,6 +260,8 @@ class BatchExperimentRun(BaseModel):
     """If the batch experiment is a scenario test, this is the ID of that test."""
     repetition: int | None = None
     """Repetition number of the experiment."""
+    run_number: str | None = None
+    """Run number of the experiment."""
```
merschformann, line 264: Huh, this was missed before, right?

nextmv/tests/cli/test_mcp.py
merschformann, line 1: Well that is one hell of a test file. 😂

merschformann (commented):
I have my concerns about the high degree of duplication that this one introduces, but it might be necessary (to some extent). I will defer to others for deciding here. 😊

ryanjoneil:
> I have my concerns about the high degree of duplication that this one introduces, but it might be necessary (to some extent). I will defer to others for deciding here. 😊

The issue is that agents currently aren't good at using the CLI. However, it might be possible to get the same result with some changes to how this is structured.

* Add `--yes` and other flags to CLI commands as necessary.
* Provide an MCP command that is just a thin wrapper to the CLI with additional context.
* Either keep MCP-specific features (caching, large file handling for context windows) in MCP or move them into CLI.

I can take a look at that.

sebastian-quintero:
image

This is not true LOL.

```bash
$ nextmv cloud ensemble -h

 Usage: nextmv cloud ensemble [OPTIONS] COMMAND [ARGS]...

 Create and manage Nextmv Cloud ensemble definitions.

 An ensemble definition defines how to coordinate and execute multiple child
 runs for an application, and how to determine the optimal result from those
 runs. You can configure run groups to specify which instances to run on and
 with what options, as well as evaluation rules to determine the best result
 based on specified metrics.

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help  -h        Show this message and exit.                                                                                                                │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ create  Create a new Nextmv Cloud ensemble definition.                                                                                                       │
│ delete  Deletes a Nextmv Cloud ensemble definition.                                                                                                          │
│ get     Get a Nextmv Cloud ensemble definition.                                                                                                              │
│ list    List all Nextmv Cloud ensemble definitions for an application.                                                                                       │
│ update  Update a Nextmv Cloud ensemble definition.                                                                                                           │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

sebastian-quintero (approved):
I did an overall read of the changes, without dwelling too deep into the AI-generated code under `cli/mcp/tools`. The overall strategy is fine by me and I like that we are placing this as an optional dependency for people who want it. The architecture and the distribution of files and logic is also good.

Here is my only suggestion:
I would like that, if possible, we get rid of the double implementation of the SDK that is happening under `tools/<acceptance|account|app|batch|...>.py`. We are basically re-declaring the SDK's methods, like this one, for example:

```python
    @mcp.tool()
    def cloud_list_acceptance_tests(app_id: str) -> list[dict[str, Any]]:
        """List acceptance tests for a Nextmv Cloud application.

        Acceptance tests compare a candidate instance against a
        baseline using defined metrics. Returns a list of test
        summaries.

        Args:
            app_id: The application ID.
        """

        app = _helpers._get_app(app_id)
        tests = app.list_acceptance_tests()
        return [t.to_dict() for t in tests]
```

What mostly gets me is the `def` line. We are already maintaining two declarations of types for every operation we want to perform: in the SDK and in the CLI. I would like for us not to have a third place where we have to maintain and control when a method gets a new argument, for example.

My ask: is there a way in which we can use the `@mcp.tool` decorator in our current methods of the SDK and `register` them in a single place? It is mostly re-declaring our SDK that I dislike.

If this is unreasonable, I am also ok with moving forward and finding out how this breaks down the road. For now, I don't have any reasons to block this. Onwards!

merschformann (approved):
🎉

ryanjoneil:
> My ask: is there a way in which we can use the `@mcp.tool` decorator in our current methods of the SDK and `register` them in a single place? It is mostly re-declaring our SDK that I dislike.

I believe so, but it will have to wait for a week or two. Merging this per discussion for the upcoming conference.

merschformann (approved):
🎉

PR #241, opened 2026-02-24
Feat/eng 7122 add sso delete domain mapping

chriskeyser:
# Description

This feature adds the SSO domain command tree. Creating an SSO domain mapping is under Nextmv control, so its in our internal `nextmv-ops tool` until we can do txt record validation for domain ownership.

* `nextmv cloud sso domain delete --domain foo.bar --profile local` - deletes an already mapped domain from an existing SSO configuration
* `nextmv sso get` now returns `mapped_domains` as a field if there are mapped domains

### DIFF
--- a/nextmv/nextmv/cli/cloud/sso/__init__.py
+++ b/nextmv/nextmv/cli/cloud/sso/__init__.py
@@ -7,6 +7,7 @@ import typer
 from nextmv.cli.cloud.sso.create import app as create_app
 from nextmv.cli.cloud.sso.delete import app as delete_app
 from nextmv.cli.cloud.sso.disable import app as disable_app
+from nextmv.cli.cloud.sso.domain import app as domain_app
 from nextmv.cli.cloud.sso.enable import app as enable_app
 from nextmv.cli.cloud.sso.get import app as get_app
 from nextmv.cli.cloud.sso.update import app as update_app
@@ -19,6 +20,7 @@ app.add_typer(disable_app)
 app.add_typer(enable_app)
 app.add_typer(get_app)
 app.add_typer(update_app)
+app.add_typer(domain_app, name="domain")
 
 
 @app.callback()
--- a/nextmv/nextmv/cli/cloud/sso/create.py
+++ b/nextmv/nextmv/cli/cloud/sso/create.py
@@ -108,5 +108,6 @@ def create(
         enabled=enabled,
         metadata_url=metadata_url,
         metadata_document=metadata_document,
+        mapped_domains=None,  # mapped_domains cannot be set at creation
     )
     success("SSO configuration created successfully.")
--- a/nextmv/nextmv/cli/cloud/sso/create.py
+++ b/nextmv/nextmv/cli/cloud/sso/create.py
@@ -108,6 +108,5 @@ def create(
         enabled=enabled,
         metadata_url=metadata_url,
         metadata_document=metadata_document,
-        mapped_domains=None,  # mapped_domains cannot be set at creation
     )
     success("SSO configuration created successfully.")
--- a/nextmv/nextmv/cli/cloud/sso/domain/__init__.py
+++ b/nextmv/nextmv/cli/cloud/sso/domain/__init__.py
@@ -1,0 +1,22 @@
+"""
+This module defines the cloud sso command tree for the Nextmv CLI.
+"""
+
+import typer
+
+from nextmv.cli.cloud.sso.domain.delete import app as delete_app
+
+# Set up subcommand application.
+app = typer.Typer()
+app.add_typer(delete_app)
+
+
+@app.callback()
+def callback() -> None:
+    """
+    Manage SSO mapped domains for your Nextmv Cloud organization (account).
+
+    Please contact [link=https://www.nextmv.io/contact][bold]Nextmv support[/bold][/link]
+    for assistance configuring SSO for your organization.
+    """
+    pass
--- a/nextmv/nextmv/cli/cloud/sso/domain/__init__.py
+++ b/nextmv/nextmv/cli/cloud/sso/domain/__init__.py
@@ -1,5 +1,5 @@
 """
-This module defines the cloud sso command tree for the Nextmv CLI.
+This module defines the cloud sso domain command tree for the Nextmv CLI.
 """
 
 import typer
@@ -15,6 +15,8 @@ app.add_typer(delete_app)
 def callback() -> None:
     """
     Manage SSO mapped domains for your Nextmv Cloud organization (account).
+
+    Mapped domains redirect additional domains to your IDP for federated authentication.
 
     Please contact [link=https://www.nextmv.io/contact][bold]Nextmv support[/bold][/link]
     for assistance configuring SSO for your organization.
--- a/nextmv/nextmv/cli/cloud/sso/domain/delete.py
+++ b/nextmv/nextmv/cli/cloud/sso/domain/delete.py
@@ -16,6 +16,15 @@ app = typer.Typer()
 
 @app.command()
 def delete(
+    domain: Annotated[
+        str,
+        typer.Option(
+            "--domain",
+            "-d",
+            help="The domain to delete from the SSO configuration.",
+            metavar="DOMAIN",
+        ),
+    ],
     yes: Annotated[
         bool,
         typer.Option(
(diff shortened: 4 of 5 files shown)


nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,67 @@
...
+        )
+
+        if not confirm:
+            info("SSO mapped domain will not be deleted.")
+            return
+
+    sso_config = build_sso_config(profile)
+    in_progress(msg="Deleting SSO mapped domain from configuration...")
+    sso_config.delete_domain(domain=domain)
+    success("SSO mapped domain deleted successfully.")
```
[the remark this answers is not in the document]
chriskeyser: will be in next update

nextmv/nextmv/cloud/sso.py
```diff
@@ -101,6 +104,11 @@ class SSOConfiguration(BaseModel):
     """
     The SSO metadata document as a string.
     """
+    mapped_domains: list[str] = Field(default_factory=list)
+    """
+    A list of mapped domains in the SSO configuration. Mapped domains redirect
+    additional domains to your IDP for federated authentication.
+    """
```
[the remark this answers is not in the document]
chriskeyser: in next update

merschformann (approved):
Looks good, thank you! 🤗

nextmv/nextmv/cli/cloud/sso/domain/__init__.py
```diff
@@ -0,0 +1,22 @@
+"""
+This module defines the cloud sso command tree for the Nextmv CLI.
```
sebastian-quintero:
```suggestion
This module defines the cloud sso domain command tree for the Nextmv CLI.
```

Missing the `domain` here.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,66 @@
+"""
+This module defines the cloud sso delete domain command for the Nextmv CLI.
```
sebastian-quintero:
```suggestion
This module defines the cloud sso domain delete command for the Nextmv CLI.
```

Changing ordering of the words so that the command reads as it is typed.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,66 @@
...
+            "-y",
+            help="Agree to deletion confirmation prompt. Useful for non-interactive sessions.",
+        ),
+    ] = False,
+    profile: ProfileOption = None,
+) -> None:
+    """
+    Delete a mapped domain from a Nextmv Cloud SSO configuration.
+
+    Mapped domains redirect additional domains to your IDP for federated authentication.
```
sebastian-quintero: I think this line belongs in the `__init__.py`, because it is the description of what a mapped domain is.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,66 @@
...
+    This command deletes a domain from an existing SSO configuration.
```
sebastian-quintero: This line is redundant I think, as the short description already states this.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,66 @@
...
+
+    You can use the command [code] nextmv cloud sso get[/code] command to view all
```
sebastian-quintero:
```suggestion
    You can use the [code]nextmv cloud sso get[/code] command to view all
```

The "command" word is repeated here and there is an unnecessary extra space after opening the `code` tag.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
```diff
@@ -0,0 +1,66 @@
...
+    mapped domains in your SSO configuration.
+
```
sebastian-quintero, line 47: See other `delete` commands and copy the description of the usage of the `--yes` flag. Something along the lines of:

```text
This action is permanent and cannot be undone. Use the --yes
 flag to skip the confirmation prompt.
```

nextmv/nextmv/cli/cloud/sso/__init__.py
```diff
@@ -19,6 +20,7 @@
 app.add_typer(enable_app)
 app.add_typer(get_app)
 app.add_typer(update_app)
+app.add_typer(domain_app, name="domain")
```
sebastian-quintero, line 23: Thinking out loud. Can the `domain` command tree belong under `cloud` directly to avoid the deep nesting? Probably not.

nextmv/nextmv/cloud/sso.py
```diff
@@ -144,6 +151,7 @@ def new(
         enabled: bool | None = None,
         metadata_url: str | None = None,
         metadata_document: str | None = None,
+        mapped_domains: list[str] | None = None,
```
sebastian-quintero: The `new` method is not going to use `mapped_domains` for anything. In the class definition, the `mapped_domains` attribute will be initialized as `None` because that is the default value. Remove `mapped_domains` as an argument to the `new` method.
```suggestion
```

nextmv/nextmv/cli/cloud/sso/create.py
```diff
@@ -108,5 +108,6 @@ def create(
         enabled=enabled,
         metadata_url=metadata_url,
         metadata_document=metadata_document,
+        mapped_domains=None,  # mapped_domains cannot be set at creation
```
sebastian-quintero: See my comment in the `new` method: you don't need `mapped_domains` as an argument to the `new` method.
```suggestion
```

nextmv/nextmv/cloud/sso.py
```diff
@@ -165,6 +173,8 @@ def new(
             The URL to the SSO metadata document.
         metadata_document : str, optional
             The SSO metadata document as a string.
+        mapped_domains : list[str], optional
+            A list of mapped domains in the SSO configuration.
```
sebastian-quintero:
```suggestion
```

Per my comment above (in the `new` method, remove this from the docstring as well.

sebastian-quintero (changes requested):
Most comments are optional except for the one relating to the `new` method, where I don't think we should add `mapped_domains` as an argument to the method.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
chriskeyser: I followed the lead of the `sso delete` command. Since it can just be reconfigured and no data loss, this seemed appropriate? I don't have an issue changing it if you want however there is no long term "damage" to do this like deleting an app or run.

nextmv/nextmv/cli/cloud/sso/__init__.py
chriskeyser: I don't think we should do that as it directly applies to SSO configuration, ie the command makes no sense outside of an SSO context and its possible we have other operations related to domains in other areas eventually (thinking external integrations of some sort). I followed the model in the contribution guideline, I originally had the command directly under sso as `delete-domain` but that felt like it wasn't properly following the model.

I'd be more concerned on depth too if the command was used frequently. This command set will be infrequently used.

I'm open to updates though if you want to push it in a certain direction.

nextmv/nextmv/cli/cloud/sso/domain/delete.py
chriskeyser: I could say something like:

*This action will prevent users in the domain {domain} from accessing your account using SSO.* if that feels better too.

chriskeyser: Pushed a second update that I think addresses your concern.

sebastian-quintero: Yeah all good, I was mainly pushing for the description to say "You can use the --yes flag...".

nextmv/nextmv/cli/cloud/sso/__init__.py
sebastian-quintero: Cool, I agree. I was merely thinking out loud, no action needed.

sebastian-quintero (approved):
Thank you for the work.

PR #243, opened 2026-02-27
Feature/eng 7077 update python sdk and cli to use managed inputs for runs

cowanwalls:
# Description

Adds managed input handling for standard flow app runs (i.e. non-experiment)

## Changes

- `nextmv/nextmv/run.py`
    - Updated run metadata to include tracking return information (`input_id` field for managed inputs and `cloned_run_id` for console flow usage)
    - Adds experiment ID and experiment type to run metadata return
- `nextmv/nextmv/cloud/application/_run.py`
    - `new_run` now handles managed input
    - `new_run_with_result` now also handles managed input
- `nextmv/nextmv/cli/cloud/run/create.py`
    - command now includes optional managed input flag.
        - Standard flag for managed inputs not used as it is not an optional by way of being an `| None` type, which is why it was not used
    - handles managed input flag
- `nextmv/tests/integration/cloud/test_integration_cloud.py`
    - url may be specified on client now
        - Allows Nextmv developers to run tests in non-production environments in a more streamlined way
        - Potentially allows for negative testing against non-production environments if we ever need to do so in the future
            - Would require more secrets managed on GH- potential security concern
            - Unlikely to actually have an impact in this area long term (automated negative integration tests of the python sdk seems unlikely)
    - Sets up and starts 2 additional runs (and expects their successful completion):
        - Run with input via upload URL
        - Run with managed input

## Testing

Manual testing of the CLI for these changes was done. Looked nominal. Integration tests ran successfully against a non-production environment.


nextmv/tests/integration/cloud/test_integration_cloud.py
```diff
@@ -8,7 +8,7 @@
 import nextmv
 from nextmv import cloud
 
-client = cloud.Client(api_key=os.getenv("NEXTMV_API_KEY"))
+client = cloud.Client(api_key=os.getenv("NEXTMV_API_KEY"), url=os.getenv("NEXTMV_BASE_URL"))
```
merschformann, line 11: We should think about redirecting testing to staging in general. 🤔

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -362,8 +366,9 @@ def new_run(  # noqa: C901 # Refactor this function at some point.
         if input_data is not None:
             input_size = get_size(input_data)
 
-        upload_id_used = upload_id is not None
-        if self.__upload_url_required(upload_id_used, input_size, tar_file, input):
+        managed_input_id_used = managed_input_id is not None and managed_input_id != ""
+        upload_id_used = upload_id is not None and upload_id != ""
+        if self.__upload_url_required(upload_id_used or managed_input_id_used, input_size, tar_file, input):
             upload_url = self.upload_url()
             self.upload_data(data=input_data, upload_url=upload_url, tar_file=tar_file)
```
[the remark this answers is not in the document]
merschformann: Not sure whether this is worth it tbf.

nextmv/nextmv/cloud/client.py
```diff
@@ -152,6 +152,10 @@ def __post_init__(self):
             selected profile.
         """
 
+        # Fallback for integration tests
+        if self.url is None:
```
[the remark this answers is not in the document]
merschformann: Fair point, and another one where `if not self.url:` would be a Pythonic shortcut. However, I am not sure whether we wouldn't want to run integration tests against staging in general. 🤷 

Not a matter for this PR though.

nextmv/nextmv/run.py
```diff
@@ -720,6 +720,30 @@ class Run(BaseModel):
...
 
+class RunTrackingMetadata(BaseModel):
+    """
+    Metadata elements for tracking elements of the run that
+    may be useful for the purposes of reproducability.
+
+    You can import the RunTrackingMetadata class directly from `nextmv`:
+
+    ```python
+    from nextmv import RunTrackingMetadata
```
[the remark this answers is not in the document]
merschformann: Yep, good catch. `RunTrackingMetadata` still needs to get exported. ☺️

nextmv/nextmv/cli/cloud/run/create.py
```diff
@@ -297,12 +308,17 @@ def create(
...
+      Wait for the run to complete and download the result files to an [magenta]outputs[/magenta] directory.
+        $ [dim]nextmv cloud run create --app-id hare-app --managed-input-id carrot-input --output outputs[/dim]
     """
 
     # Validate that input is provided.
     stdin = sys.stdin.read().strip() if sys.stdin.isatty() is False else None
-    if stdin is None and (input is None or input == ""):
-        error("Input data must be provided via the --input flag or [magenta]stdin[/magenta].")
+    if stdin is None and (input is None or input == "") and (managed_input_id is None or managed_input_id == ""):
+        error("Input data must be provided via the --input or --managed_input_id flags, or [magenta]stdin[/magenta].")
```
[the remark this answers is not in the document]
merschformann: Good catch.

nextmv/tests/integration/cloud/test_integration_cloud.py
```diff
@@ -247,7 +256,7 @@ def runs(app: cloud.Application, instances: tuple[cloud.Instance, cloud.Instance
 
         # We can list runs.
         run_list = app.list_runs()
-        assert len(run_list) >= 5
+        assert len(run_list) >= 7
```
merschformann, line 259: I know it has been here already, but don't we know the exact number of runs? Why are we checking `ge` here? Or, is it to avoid issues when testing locally and the app didn't get cleaned up? 

Just curious though, nothing to fix here. ☺️

merschformann (approved):
Looks good, the bots had some good but minor comments. ☺️

nextmv/nextmv/cloud/client.py
```diff
@@ -153,7 +158,7 @@ def __post_init__(self):
         """
 
         # Fallback for integration tests
-        if self.url is None:
+        if not self.url :
```
merschformann:
```suggestion
        if not self.url:
```

I think the formatter (ruff) didn't run here. Just a tiny nit, I hope the suggestion helps. ☺️

merschformann (approved):
Thank you! 🤗 

Just noted a tiny formatting issue. 😊

nextmv/nextmv/cloud/client.py
```diff
@@ -152,6 +157,10 @@ def __post_init__(self):
             selected profile.
         """
 
+        # Fallback for integration tests
+        if not self.url:
+            self.url = "https://api.cloud.nextmv.io"
```
sebastian-quintero, line 162: This shouldn't be necessary, because line 131 already sets the default when not provided.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -239,6 +239,7 @@ def list_runs(self: "Application", status: StatusV2 | None = None) -> list[Run]:
     def new_run(  # noqa: C901 # Refactor this function at some point.
         self: "Application",
         input: Input | dict[str, Any] | BaseModel | str = None,
+        managed_input_id: str | None = None,
```
sebastian-quintero, line 242: This argument should be the last one, otherwise this is a breaking change for people using this call with positional parameters.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -287,6 +288,9 @@ def new_run(  # noqa: C901 # Refactor this function at some point.
 
             In general, if an input is too large, it will be uploaded with the
             `upload_data` method.
+        managed_input_id: Optional[str]
+            The ID of an existing managed input (`nextmv.cloud.ManagedInput`)
+            to use as the run's input.
```
sebastian-quintero, line 293: Move here too, to be the last argument.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -362,8 +366,9 @@ def new_run(  # noqa: C901 # Refactor this function at some point.
         if input_data is not None:
             input_size = get_size(input_data)
 
-        upload_id_used = upload_id is not None
-        if self.__upload_url_required(upload_id_used, input_size, tar_file, input):
+        managed_input_id_used = managed_input_id is not None and managed_input_id != ""
+        upload_id_used = upload_id is not None and upload_id != ""
+        if self.__upload_url_required(upload_id_used or managed_input_id_used, input_size, tar_file, input):
```
sebastian-quintero, line 371: Use keyword arguments here for more clarity, as opposed to arguments directly.
```suggestion
        if self.__upload_url_required(
            upload_id_used=upload_id_used or managed_input_id_used,
            input_size=input_size,
            tar_file=tar_file,
            input=input,
        ):
```

sebastian-quintero (commented):
@cowanwalls sorry for the late review. We should at a minimum adjust the argument in the `_run.py` file to avoid incurring into a breaking change.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -417,6 +431,7 @@ def new_run(  # noqa: C901 # Refactor this function at some point.
     def new_run_with_result(
         self: "Application",
         input: Input | dict[str, Any] | BaseModel | str = None,
+        managed_input_id: str | None = None,
```
sebastian-quintero, line 434: Same here, should be the last argument.

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -522,6 +537,8 @@ def new_run_with_result(
             Path to a directory where non-JSON output files will be saved. This is
             required if the output is non-JSON. If the directory does not exist, it
             will be created. Uses the current directory by default.
+        managed_input_id : Optional[str],
+            The ID of an existing managed input (`nextmv.cloud.ManagedInput`) to use as the run's input.
```
sebastian-quintero, line 541: I think this is already documented as the last arg.

PR #244, opened 2026-03-02
addresses breaking changes

cowanwalls:
# Description

Addresses feedback from [prior PR](https://github.com/nextmv-io/nextmv-py/pull/243) to avoid breaking changes

## Changes

- `nextmv/nextmv/cloud/application/_run.py`
    - ensures new function arguments are at the end to avoid the function signature breaking for anyone not using keyword args up to at least `instance_id`
    - When checking if a run is made via an uploaded input, use `kwargs` for visual clarity
- `nextmv/nextmv/cloud/client.py`
    - Actual important part of `if not self.url` check commented in now (does not introduce breaking changes)
    
## Testing

Tried to be thorough with this but VSCode and `debugpy` were not cooperating for me so had to punt for the sake of getting changes out in a reasonable time.


sebastian-quintero (approved):
Thank you for addressing 🌮

PR #246, opened 2026-03-03
Chore/integ test stability improvements

cowanwalls:
# Description

Updates integration tests for increased stability

## Changes

- `nextmv/tests/integration/cloud/test_integration_cloud.py`
    - Adds `TODO` statements for making a bespoke python integration test community app
    - Comments out cancel run test
        - This was running on a `hello world` app that completed quickly, so the test was ineffective at testing cancellation reliably
    - Waits for scenario test to complete before deleting it
    - Expects switchback and shadow tests to be in a final or stopping state after stop endpoint is called
    - Expects ensemble run to complete and succeed
    - Waits for 1 second before deleting the app to allow prior delete operations to finalize
- `nextmv/tests/integration/cloud/test_integration_marketplace.py`
    - Adds base URL to client construction (for non-prod testing)
    
## Testing

Integration tests ran against non-prod environment without issues.


sebastian-quintero (approved):
Ok hopefully we don't forget about those TODOs 😅

nextmv/tests/integration/cloud/test_integration_cloud.py
```diff
@@ -744,6 +752,7 @@ def ensembles(app: cloud.Application, instances: tuple[cloud.Instance, cloud.Ins
                 ),
             ),
         )
+        assert run.metadata.status_v2 == nextmv.StatusV2.succeeded
```
merschformann, line 755: I wanted to add a suggestion here to maybe rename run to result to be a bit more on point, but GH UI is apparently a bit broken with suggestions atm. Anyway, fine to ignore. 😊

merschformann (approved):
👍

nextmv/tests/integration/cloud/test_integration_cloud.py
cowanwalls: Agreed this should get changed, but want to prioritize getting this merged so we can get Python development unmoored. I can pick this up in a docs PR I have coming in a few minutes after this.

PR #256, opened 2026-03-25
Fix bug where app.yaml file was re-formatted when getting a registered local app

sebastian-quintero:
# Description

When running an app locally, and the app was being fetched from the local registry, the `app.yaml` was being overwritten with default `.yaml` formatting, getting rid of comments and the user's format. Although the contents remain unchanged, the format changed entirely, which is undesirable.


merschformann (approved):
🎉

PR #263, opened 2026-04-02
Improve `bash` detection from `_get_shell_command_elements`

### LINKED ISSUE #262: Pre-push command fails to recognize command
I am using the nextmv v1.4.0 package's CLI from Git Bash on a Windows 11 laptop. My nextmv app has a pre-push command to export the requirements of the app prior to being pushed to nextmv.

```yaml
type: python
runtime: ghcr.io/nextmv-io/runtime/python:3.11
pre-push: |
  uv export \
    --project . \
    --format requirements-txt \
    --output-file requirements.txt \
    --no-hashes \
    --no-dev \
    --no-annotate \
    --no-emit-project \
    --no-emit-workspace \
    --no-editable
python:
  pip-requirements: requirements.txt
```

When I run `nextmv cloud app push` I get:

```
Error: error running pre-push command: /bin/bash: line 1: uv: command not found.
```

```
$ nextmv cloud app push -a routing-pipeline
💿 Starting build for Nextmv application routing-pipeline.
🔨 Running pre-push command: bash -c uv export \
  --project . \
  --format requirements-txt \
  --output-file requirements.txt \
  --no-hashes \
  --no-dev \
  --no-annotate \
  --no-emit-project \
  --no-emit-workspace \
  --no-editable

Error: error running pre-push command: /bin/bash: line 1: uv: command not found.
```

Without explicitly detecting the correct `bash`, I believe `subprocess.run` would resolve `bash` from the system `PATH`, which might not be desirable.

```
pryerc MINGW64 ~/github/twin-labs/twin2/routing-pipeline (chris/nextpipe)
$ where bash
C:\Users\pryerc\AppData\Local\Programs\Git\usr\bin\bash.exe
C:\Windows\System32\bash.exe
C:\Users\pryerc\AppData\Local\Microsoft\WindowsApps\bash.exe

pryerc MINGW64 ~/github/twin-labs/twin2/routing-pipeline (chris/nextpipe)
$ C:\\Windows\\System32\\bash.exe -c 'uv'
/bin/bash: line 1: uv: command not found

pryerc MINGW64 ~/github/twin-labs/twin2/routing-pipeline (chris/nextpipe)
$ C:\\Users\\pryerc\\AppData\\Local\\Microsoft\\WindowsApps\\bash.exe -c 'uv'
/bin/bash: line 1: uv: command not found

pryerc MINGW64 ~/github/twin-labs/twin2/routing-pipeline (chris/nextpipe)
$ C:\\Users\\pryerc\\AppData\\Local\\Programs\\Git\\usr\\bin\\bash.exe -c 'uv --version'
uv 0.11.3 (45da18ac3 2026-04-01 x86_64-pc-windows-msvc)
```

cnpryer:
# Changes

Previously, if `SHELL` is set and `shutil.which("bash")` exists `bash` would be used as the executable. 

This PR modifies `_get_shell_command_elements` to use `shutil.which("bash")` itself as the executable to ensure the correct `bash` is used from a separate process. 

This assumes that `SHELL` points to `bash`, but that is an assumption already made in the original code.

# Testing

I made the changes to `nextmv` locally and ran:

```
uvx --from <path> nextmv cloud app push
```

I also ran:

```
uv run pytest tests/test_package.py
```

And 

```
uv run pytest tests/integration -s
```

Closes #262

### DIFF
--- a/nextmv/nextmv/cloud/package.py
+++ b/nextmv/nextmv/cloud/package.py
@@ -134,8 +134,9 @@ def _run_build_command(
 def _get_shell_command_elements(pre_push_command):
     """Get the shell command elements based on the operating system."""
     # Check if we're in a Unix-like shell (including MINGW on Windows)
-    if "SHELL" in os.environ and shutil.which("bash"):
-        return ["bash", "-c", pre_push_command]
+    bash = shutil.which("bash")
+    if "SHELL" in os.environ and bash:
+        return [bash, "-c", pre_push_command]
     # Default to cmd on Windows
     elif platform.system() == "Windows":
         return ["cmd", "/c", pre_push_command]
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -1,5 +1,11 @@
 name: python lint
-on: [push]
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+permissions:
+  contents: read
 jobs:
   python-lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,5 +1,11 @@
 name: unit tests
-on: [push]
+on:
+  push:
+    branches: [develop]
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+permissions:
+  contents: read
 jobs:
   test:
     runs-on: ${{ matrix.platform }}


nextmv/nextmv/cloud/package.py
```diff
@@ -133,9 +133,10 @@ def _run_build_command(
 
 def _get_shell_command_elements(pre_push_command):
     """Get the shell command elements based on the operating system."""
-    # Check if we're in a Unix-like shell (including MINGW on Windows)
-    if "SHELL" in os.environ and shutil.which("bash"):
-        return ["bash", "-c", pre_push_command]
+    # Check if we're in a Unix-like shell and use `bash` (including MINGW on Windows)
+    shell = os.environ.get("SHELL")
+    if shell and shell in shutil.which("bash"):
+        return [shell, "-c", pre_push_command]
```
cnpryer: A few problems with this. 

First, I was mixing my `where` with `shutil.which`. `shutil.which` will return a string of the path to `bash` if it exists. And the string returned isn't guaranteed to be an exact match. If an exact match is what you'd like here then we can evaluate with `os.path.abspath` equality. But maybe that's not a good idea.

Requiring an exact match without another branch for the if-statement might cause a breaking change in the behavior of this function in some environments. If `SHELL` is not pointed to `bash` (for some reason), then it will no longer just defer to the resolved `bash` if there is one.

Otherwise the alternative is to just use `shutil.which("bash")` since the idea originally was to just defer to `bash` anyway. Let me know if you have different thoughts. Happy to make adjustments.

cnpryer: https://github.com/nextmv-io/nextmv-py/pull/263/changes/c2769f7e07bea6bad95fa878de99df928bcbfacb

nextmv/nextmv/cloud/package.py
```diff
@@ -134,8 +134,9 @@ def _run_build_command(
 def _get_shell_command_elements(pre_push_command):
     """Get the shell command elements based on the operating system."""
     # Check if we're in a Unix-like shell (including MINGW on Windows)
-    if "SHELL" in os.environ and shutil.which("bash"):
-        return ["bash", "-c", pre_push_command]
+    bash = shutil.which("bash")
+    if "SHELL" in os.environ and bash:
```
cnpryer, line 138: Could also use the walrus operator.

```suggestion
    if "SHELL" in os.environ and (bash := shutil.which("bash")):
```

merschformann: I think the first version will be considered more intuitive, though, I wouldn't mind the walrus personally. ☺️ 

Image

merschformann (approved):
Ah, this is probably about the `.exe` extension added by Windows, right? So, before invoking `bash` didn't fire as `bash.exe` was expected?

Otherwise, I fail to see the difference between the two. The new one respects updates to the `PATH` that the app did, but it shouldn't have done that (and if so it shouldn't be helpful). 🤔 

Anyway, this shouldn't hurt either, and might be slightly more stable (due to mentioned OS-awareness of shutil.which that might otherwise not be given. ☺️

cnpryer:
> Ah, this is probably about the `.exe` extension added by Windows, right? So, before invoking `bash` didn't fire as `bash.exe` was expected?
> 
> Otherwise, I fail to see the difference between the two. The new one respects updates to the `PATH` that the app did, but it shouldn't have done that (and if so it shouldn't be helpful). 🤔

I think it's simply due to system environment restrictions. 

I can configure my environment (which lately has been from my personal Git Bash installation), and I've configured my `.bashrc` file appropriately. So if you're searching for `bash` from `PATH` outside my shell it will have the system-scoped installations (and the Micorsoft store proxy). Both of which are problematic for users since configuration and usage might be limited. It's just strict security policies getting in the way.

I'm happy to show you today if you'd like.

merschformann:
For anyone reading along, the emulated Linux process did not pass the full environment on to the Windows subprocess it was calling into. So, the `bash.exe` path that shutil was able to found was lost when resolving just `"bash"`. :relaxed: 

Thanks again for reporting and providing the fix! :hugs:

cnpryer:
> For anyone reading along, the emulated Linux process did not pass the full environment on to the Windows subprocess it was calling into. So, the `bash.exe` path that shutil was able to found was lost when resolving just `"bash"`. ☺️
> 
> Thanks again for reporting and providing the fix! 🤗

So I was actually curious of your `.exe` comment after our chat. I think it's actually more annoying than that lol. It was bothering me that I didn't really understand what was happening.

Recently I installed WSL. I'm not familiar with WSL. I've had very limited experience with it. `C:\Windows\System32\bash.exe` is from the WSL installation. Your comment about the `.exe` extension being a possible issue is a problem from WSL if you're using Windows installed executables. 

```
C:\Users\pryerc>C:\Windows\System32\bash.exe

pryerc@-:/mnt/c/Users/pryerc$ uv --version
Command 'uv' not found, but can be installed with:
sudo snap install astral-uv

pryerc@NFI93771:/mnt/c/Users/pryerc$ uv.exe --version
uv 0.11.3 (45da18ac3 2026-04-01 x86_64-pc-windows-msvc)
```

WSL won't translate or use an alias. Makes sense since it's not an emulator. It's a subsystem (for Linux). That's likely where you ran into the `.exe` issue before. The original error I ran into _was_ because it wasn't `uv.exe` from the WSL shell.

From `MINGW64` there's no need to use a `.exe` extension even if you're using a Windows installed executable. The emulator handles that.

```
pryerc@- MINGW64 ~
$ uv --version
uv 0.11.3 (45da18ac3 2026-04-01 x86_64-pc-windows-msvc)
```

So two takeaways:

1. Using `shutil.which("bash")` directly will avoid this altogether.
2. Even though the incorrect `bash` was used, `uv.exe` would have worked from it. But.. that's no good lol.

PR #266, opened 2026-04-06
Allows override of endpoint via env var

merschformann:
# Description

Allows override of the endpoint via `NEXTMV_ENDPOINT` env var. This is mostly useful internally or with custom deployments (it's why the flag stays hidden to avoid unnecessary confusion).

## Changes

- Allows override of the endpoint via `NEXTMV_ENDPOINT` env var.


sebastian-quintero (approved):

sebastian-quintero:
Closing in favor of #267, which is a more comprehensive implementation of the `NEXTMV_ENDPOINT` feature.

PR #285, opened 2026-04-14
Warn the user with conflicting profile and API key/endpoint configs

sebastian-quintero:
# Description

Adds new logic to the main callback of the CLI to analyze the state of profiles, API key override and endpoint override. If a conflicting configuration is found, a warning is emitted.

Worst case scenario:

```bash
$ NEXTMV_PROFILE=carrot NEXTMV_API_KEY=$DIFFERENT_NEXTMV_API_KEY NEXTMV_ENDPOINT=api.different.endpoint nextmv cloud app list --profile lettuce
🚧  Warning: Both the NEXTMV_PROFILE environment variable and the --profile/-p option are set with different values. This may cause an unexpected behavior. The env var takes precedence.
🚧  Warning: The NEXTMV_API_KEY environment variable is set, and its value differs from the API key of the carrot profile. This may cause an unexpected behavior. The env var takes precedence.
🚧  Warning: The NEXTMV_ENDPOINT environment variable is set, and its value differs from the endpoint of the carrot profile. This may cause an unexpected behavior. The env var takes precedence.
⏳ Listing applications...
```

### DIFF
--- a/nextmv/nextmv/cli/main.py
+++ b/nextmv/nextmv/cli/main.py
@@ -13,6 +13,7 @@ about the features used here. An example of Rich markup can be found in the
 epilog of the Typer application defined below.
 """
 
+import os
 import sys
 from typing import Annotated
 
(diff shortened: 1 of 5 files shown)


merschformann (approved):
Nice, also nice cleanup! Thanks! 🤗

PR #286, opened 2026-04-14
Use instance created during `init` for run execution

sebastian-quintero:
# Description

Uses the instance that is created during the `init` process to start the remote run, as opposed to always using `latest`.

## Examples

1. Created the `prod` instance, and it is used to create the run.

   image

1. Didn't create a version or instance, `latest` is used as the default.

   image

1. Created a version but didn't create an instance, `latest` is used as the default.

   image

1. _Updated_ the `prod` instance, and it is used to create the run.

   image

### DIFF
--- a/nextmv/nextmv/cli/init.py
+++ b/nextmv/nextmv/cli/init.py
@@ -118,11 +118,11 @@ def init() -> None:
             commands.extend(cmds4)
             rule()
 
-            cmd5 = _handle_app_push(cloud_app)
+            instance_id, cmd5 = _handle_app_push(cloud_app)
             commands.append(cmd5)
             rule()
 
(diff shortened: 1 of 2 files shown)


nextmv/nextmv/cli/init.py
```diff
@@ -815,6 +820,9 @@ def _handle_cloud_run_create(
 
     # Actually execute the command to start the Cloud run.
     command = ["nextmv", "cloud", "run", "create", "--app-id", cloud_app.id, "--input", dirpath]
+    if instance_id:
```
merschformann, line 823: ❤️

merschformann (approved):
🎉

PR #287, opened 2026-04-14
Remove `pip install` suggestion during `init`

sebastian-quintero:
# Description

Given that we now rely on `uv` for local run execution of `python` apps, we don't need to suggest installing deps anymore during `nextmv init`.


merschformann (approved):
🪦

PR #288, opened 2026-04-14
Avoid showing assets for JSON runs to keep terminal more compact

sebastian-quintero:
# Description

When using `nextmv cloud/local run create/get` with a `json` app that produces assets, the output typically showed the full JSON result with the assets bundled in the response:

image

This doesn't produce much value, as it is mostly illegible information.

Now, a `json` output containing assets is trimmed by default. People can still use `--output` to save the full output to a file.

image


merschformann (approved):
Good idea, thanks! ❤️

PR #296, opened 2026-04-23
Remove unnecessary Output validation

sebastian-quintero:
# Description

The `Output.output_format` was being set by default to `ContentFormat.JSON`. This is not accurate, as the changes introduced to `nextmv.write` imply that one can specify `solution_files` in the `Output` without ever specifying the format, counting on it getting picked up by the manifest.


merschformann (approved):
🎉

PR #315, opened 2026-05-13
Caches Python dependencies for pushing, adds `nextmv cache` command tree in CLI, `cache.py` module to SDK

sebastian-quintero:
# Description

This PR is about introducing a cache for storing Python dependencies for applications. The intent is that when we are re-pushing an app whose dependencies don't change, we don't have to re-download and compress all dependencies.

- Current behavior:
  - Dependencies are read from `pyproject.toml` or `requirements.txt`.
  - They are downloaded from the appropriate package index.
  - They are bundled with the application files (what is listed in the `app.yaml`), and everything is tarred and compressed.
  - Whenever we push again, the whole process is repeated.
- New behavior:
  - Dependencies are read from `pyproject.toml` or `requirements.txt`.
  - A lockfile is generated (with `uv pip`) to create precise requirements for the application.
  - For each dependency of the lockfile:
    - Compile the cache key from the name, version, python version, and platform.
    - Check if the key is already in the cache. If it is, return the artifact (more on this below). If it isn't, mark the dep as it will be needed for downloading.
  - All dependencies that were marked are now downloaded from the appropriate package index.
  - For each dependency, an artifact is generated. The artifact is a g-zipped tarball.
  - The artifact, along with metadata, is saved to the cache for future use.
  - All artifacts (cached + newly-built) are concatenated together using tar operations.
  - The app files (what is listed in the `app.yaml`) are added to the dependency compressed tarball.
  - Whenever we push again, we go through the cache again. That means that if a dep is found in the cache, it is used.

The cache is evicted with a LRU (least recently used) policy, and sensible defaults of 2000 elements / 5 GiB are used. The cache is evicted whenever new dependencies are added to the cache.  Note that the cache work on a dependency by dependency basis, meaning that if a single dependency changes, we don't need to re-download everything, just the one that changed.

Additionally:
- The `cache.py` module was added to the SDK exposing cache functionalities.
- The `nextmv cache delete` and `nextmv cache get` commands were introduced to the CLI so that users may inspect and delete the cache.
- The `--no-cache` option is introduced to `nextmv cloud app push` so that people can choose to push an application without relying on the cache. The corresponding SDK option is `no_cache` in the `cloud.Application.push` method.

## Demo

To demonstrate the functionality of the cache, _the actual push to `cloud`_ was disabled. That way we are only comparing the download and compression operations.

1. Starting with a fresh cache, push the `python-ampl-facilitylocation` app to the cloud.

image

About 56 seconds total.

2. Let's simply run the command again.

image

About 2.8 seconds total. Blazing fast ⚡.

3. What happens if I change a single dependency? In this case, I am upgrading `pandas` to the latest version.

image

About 6.1 seconds total, pretty good and _way_ faster than doing all the deps.

4. And if I run the same command again with the latest `pandas` version?

image

Everything was loaded from the cache, nice!

5. I changed my mind, let me go back to the original `pandas` version, before the upgrade.

image

Cool, as expected, everything is loaded from the cache.

6. For comparison, I switched to the `develop` branch. Let's run the same process using the old logic.

image

About 46.7 seconds. The new process, starting fresh, is _slower_ because of the individual compression of the dependencies. But that is a tradeoff that is worth it given the speedup that is gained is subsequent pushes.

### DIFF
--- a/nextmv/nextmv/cache.py
+++ b/nextmv/nextmv/cache.py
@@ -33,6 +33,8 @@ _DEPS_CACHE_DIR = _CACHE_DIR / "deps"
 """Directory storing per-package dependency tarballs."""
 _CACHE_INFO_FILE = "cache_info.json"
 """Name of the metadata file stored alongside each cached entry."""
+_DEP_TARBALL_NAME = "installed.tar.gz"
+"""Name of the per-package tarball containing installed files for a single package."""
 _MAX_DEPS = 2000
 """Maximum number of individual package tarballs to keep in the cache."""
 _MAX_DEP_BYTES = 5 * 1024**3
@@ -63,6 +65,12 @@ def get_cache() -> dict[str, Any]:
                 with open(info_file) as f:
                     info = json.load(f)
 
+                tarball = entry_dir / _DEP_TARBALL_NAME
+                try:
+                    tarball_size = format_bytes(tarball.stat().st_size)
+                except OSError:
+                    tarball_size = ""
+
                 dependencies.append(
                     {
                         "key": entry_dir.name,
@@ -71,13 +79,14 @@ def get_cache() -> dict[str, Any]:
                         "last_used_at": info.get("last_used_at", ""),
                         "python_version": info.get("python_version", ""),
                         "platform": info.get("platform", ""),
+                        "size": tarball_size,
                     }
                 )
             except Exception:
                 pass
 
     if dependencies:
-        dependencies.sort(key=lambda d: (d["last_used_at"], d["name"], d["version"]))
+        dependencies.sort(key=lambda d: (d["last_used_at"], d["name"], d["version"]), reverse=True)
 
     return {
         "num_dependencies": num_deps,
@@ -100,10 +109,7 @@ def clear_cache() -> tuple[int, int]:
         return 0, 0
 
     num_deps, total_bytes = _cache_stats()
-
-    if _CACHE_DIR.exists():
-        shutil.rmtree(str(_CACHE_DIR))
-
+    shutil.rmtree(str(_CACHE_DIR))
     _create_cache()
 
     return num_deps, total_bytes
@@ -165,7 +171,7 @@ def get_cached_dep(pkg_key: str) -> Path | None:
 
     entry_dir = _DEPS_CACHE_DIR / pkg_key
     info_file = entry_dir / _CACHE_INFO_FILE
-    installed_tar = entry_dir / "installed.tar.gz"
+    installed_tar = entry_dir / _DEP_TARBALL_NAME
 
     if not installed_tar.is_file() or not info_file.is_file():
         return None
@@ -225,7 +231,7 @@ def store_dep(
 
     with tempfile.TemporaryDirectory(dir=_DEPS_CACHE_DIR, prefix=f"{pkg_key}-tmp-", ignore_cleanup_errors=True) as _tmp:
         tmp_entry = Path(_tmp)
-        shutil.copy2(str(tar_path), tmp_entry / "installed.tar.gz")
+        shutil.copy2(str(tar_path), tmp_entry / _DEP_TARBALL_NAME)
         now = _now_iso()
         info = {
             "created_at": now,
@@ -300,6 +306,11 @@ def _create_cache() -> None:
 
     _DEPS_CACHE_DIR.mkdir(parents=True, exist_ok=True)
 
+    # Remove leftover staging dirs from previous crashed/killed runs.
+    for entry in _DEPS_CACHE_DIR.iterdir():
+        if entry.is_dir() and "-tmp-" in entry.name:
+            shutil.rmtree(str(entry), ignore_errors=True)
+
 
 def _evict_lru(max_entries: int = _MAX_DEPS, max_bytes: int = _MAX_DEP_BYTES) -> int:
     """
@@ -324,7 +335,7 @@ def _evict_lru(max_entries: int = _MAX_DEPS, max_bytes: int = _MAX_DEP_BYTES) ->
 
     entries = []
     for entry_dir in _DEPS_CACHE_DIR.iterdir():
-        if not entry_dir.is_dir():
+        if not entry_dir.is_dir() or "-tmp-" in entry_dir.name:
             continue
 
         info_file = entry_dir / _CACHE_INFO_FILE
--- a/nextmv/nextmv/cache.py
+++ b/nextmv/nextmv/cache.py
@@ -21,6 +21,7 @@ import hashlib
 import json
 import os
 import shutil
+import tarfile
 import tempfile
 from datetime import datetime, timezone
 from pathlib import Path
@@ -96,10 +97,10 @@ def get_cached_deps(key: str) -> Path | None:
     """
 
     entry_dir = _DEPS_CACHE_DIR / key
-    deps_dir = entry_dir / "deps"
+    deps_tar = entry_dir / "deps.tar.gz"
     info_file = entry_dir / _CACHE_INFO_FILE
 
-    if not deps_dir.is_dir() or not info_file.is_file():
+    if not deps_tar.is_file() or not info_file.is_file():
         return None
 
     # Update last_used_at for true LRU tracking.
@@ -113,7 +114,7 @@ def get_cached_deps(key: str) -> Path | None:
         # A failure to update the timestamp is non-fatal; still serve the hit.
         pass
 
-    return deps_dir
+    return deps_tar
 
 
 def store_deps(
@@ -162,8 +163,13 @@ def store_deps(
     # the directory no longer exists when the context manager tries to clean up.
     with tempfile.TemporaryDirectory(dir=_DEPS_CACHE_DIR, prefix=f"{key}-tmp-", ignore_cleanup_errors=True) as _tmp:
         tmp_entry = Path(_tmp)
-        tmp_deps = tmp_entry / "deps"
-        shutil.copytree(str(installed_deps_dir), str(tmp_deps))
+        tmp_deps_tar = tmp_entry / "deps.tar.gz"
+        dep_arcname = os.path.join(".nextmv", "python", "deps")
+        num_files = 0
+        with tarfile.open(str(tmp_deps_tar), "w:gz") as tar:
+            tar.add(str(installed_deps_dir), arcname=dep_arcname)
+            num_files = len(tar.getmembers())
+
         now = _now_iso()
         info = {
             "created_at": now,
@@ -171,6 +177,7 @@ def store_deps(
             "python_version": python_version,
             "platform": platform,
             "lockfile": lockfile_content,
+            "num_files": num_files,
         }
         metadata_path = tmp_entry / _CACHE_INFO_FILE
         _write_json_atomic(path=metadata_path, data=info)
--- a/nextmv/nextmv/cache.py
+++ b/nextmv/nextmv/cache.py
@@ -3,14 +3,16 @@ Module with the cache interface and operations for working with dependencies.
 
 Functions
 ---------
+get_cache
+    Get general information about the cache, including number of dependencies and total size.
+clear_cache
+    Delete the entire cache directory and recreate it empty.
 dep_cache_key
     Return a SHA-256 hex digest that uniquely identifies a single package.
 get_cached_dep
     Return the path to the cached per-package `installed.tar.gz` on a hit.
 store_dep
     Store a per-package `installed.tar.gz` into the cache and run LRU eviction.
-clear_cache
-    Delete the entire cache directory and recreate it empty.
 format_bytes
     Return a human-friendly string representation of a byte count.
 """
@@ -23,6 +25,7 @@ import shutil
 import tempfile
 from datetime import datetime, timezone
 from pathlib import Path
+from typing import Any
 
 _CACHE_DIR = Path.home() / ".nextmv" / "cache"
 """Root directory for all Nextmv cache data."""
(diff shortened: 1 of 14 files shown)


sebastian-quintero:
CoPilot didn't leave any comments. It either thinks the code is perfect or the code is so messed up that it didn't even bother trying to comprehend it LOL.

merschformann:
> CoPilot didn't leave any comments. It either thinks the code is perfect or the code is so messed up that it didn't even bother trying to comprehend it LOL.

It says:

> No critical bugs found overall, but given how complex this is, I'm marking it as too intricate for automated approval.

Here's hoping that I have more trust in myself than Copilot. :sweat_smile:

nextmv/nextmv/cli/cache/delete.py
```diff
@@ -0,0 +1,48 @@
...
+            "Are you sure you want to delete the Nextmv cache? This action cannot be undone.",
+        )
+
+        if not confirm:
+            info("The Nextmv cache will not be deleted.")
+            return
+
+    size_str = format_bytes(bytes_deleted)
+    dep_label = "dependency" if deps == 1 else "dependencies"
+    success(f"Deleted [magenta]{deps}[/magenta] {dep_label} ([magenta]{size_str}[/magenta]) from the cache.")
```
merschformann, line 48: I like that we are feedbacking what size we are freeing here. ❤️

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+                        "last_used_at": info.get("last_used_at", ""),
+                        "python_version": info.get("python_version", ""),
+                        "platform": info.get("platform", ""),
+                    }
+                )
+            except Exception:
+                pass
+
+    if dependencies:
+        dependencies.sort(key=lambda d: (d["last_used_at"], d["name"], d["version"]))
```
merschformann: The dependencies list is sorted ascending by `last_used_at` (oldest first). It feels like for a user most recently used dependencies could be more interesting. But just a thought. Maybe that's not the case. 🤷

```suggestion
        dependencies.sort(key=lambda d: (d["last_used_at"], d["name"], d["version"]), reverse=True)
```

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+                    info = json.load(f)
+
+                dependencies.append(
+                    {
+                        "key": entry_dir.name,
+                        "name": info.get("name", ""),
+                        "version": info.get("version", ""),
+                        "last_used_at": info.get("last_used_at", ""),
+                        "python_version": info.get("python_version", ""),
+                        "platform": info.get("platform", ""),
```
merschformann, line 81: Can we have size per dependency too? (instead of just total) 🥺 

Only if its easy though.

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+        A tuple of (num_deps_deleted, total_bytes_deleted).
+    """
+
+    if not _CACHE_DIR.exists():
+        return 0, 0
+
+    num_deps, total_bytes = _cache_stats()
+
+    if _CACHE_DIR.exists():
+        shutil.rmtree(str(_CACHE_DIR))
```
merschformann: Isn't `_CACHE_DIR.exists()` redundant after we invoked it just 3 lines above basically? 🤔 
Doesn't hurt though, so, feel free to ignore. 😊 

```suggestion
    shutil.rmtree(str(_CACHE_DIR))
```

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+    max_entries : int, optional
+        Maximum number of package tarball entries to retain.
+    max_bytes : int, optional
+        Maximum total cache size in bytes.
+    """
+
+    _create_cache()
+    entry_dir = _DEPS_CACHE_DIR / pkg_key
+
+    with tempfile.TemporaryDirectory(dir=_DEPS_CACHE_DIR, prefix=f"{pkg_key}-tmp-", ignore_cleanup_errors=True) as _tmp:
```
merschformann, line 232: I might be missing something, but I would really like not to create temp dirs inside of the `~/.nextmv/` dir. We would run the risk of directory litter whenever CLI gets killed, canceled, etc. in an unfortunate moment. Isn't it possible to do the work in a proper (`/tmp/...`) dir and just move it out via `tmp_entry.rename(entry_dir)`? 🤔 

```suggestion
    with tempfile.TemporaryDirectory(prefix=f"{pkg_key}-tmp-", ignore_cleanup_errors=True) as _tmp:
```

☝️ not necessarily the full fix, just for inspiration. 😅

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+    -------
+    int
+        The number of entries that were removed.
+    """
+
+    if not _DEPS_CACHE_DIR.is_dir():
+        return 0
+
+    entries = []
+    for entry_dir in _DEPS_CACHE_DIR.iterdir():
```
merschformann, line 337: This would also include the aforementioned temp dirs, if they are put under `_DEPS_CACHE_DIR`. So, avoiding (concurrency) problems here would be another benefit of moving to a proper temp dir.

nextmv/nextmv/cache.py
```diff
@@ -0,0 +1,420 @@
...
+
+    Parameters
+    ----------
+    path : Path
+        The path to write the JSON data to.
+    data : dict
+        The data to write as JSON.
+    """
+
+    with tempfile.NamedTemporaryFile(mode="w", dir=path.parent, prefix=".tmp-", delete=False) as f:
```
merschformann, line 423: Why is this also defining a dir? Is it because it is likely the same hardware disk? And assuming that `/tmp/` might be on a different disk and hence slow during move? I am not sure whether that is worth a worry. 🤔 ... Given the other problems this may cause.

```suggestion
    with tempfile.NamedTemporaryFile(mode="w", prefix=".tmp-", delete=False) as f:
```

nextmv/nextmv/cloud/package.py
```diff
@@ -186,20 +220,204 @@ def _run_pre_push_command(
...
+        try:
+            os.makedirs(target_dir, exist_ok=True)
+        except OSError as e:
+            raise Exception(f"error creating directory for asset file {file['interior_path']}: {e}") from e
 
-def __handle_python(
-    app_dir: str,
+        try:
+            shutil.copy2(file["absolute_path"], os.path.join(temp_dir, target_dir))
+        except subprocess.CalledProcessError as e:
```
merschformann: I think this might be the wrong error(s) to catch for `shutil.copy2`. Could you confirm? Just to make sure we are not missing the descriptive error message below.

nextmv/nextmv/cloud/package.py
```diff
@@ -212,110 +430,778 @@ def __handle_python(
...
+        tar_path = Path(out_dir) / f"{safe_name}.tar.gz"
+        dep_arcname = os.path.join(".nextmv", "python", "deps")
+
+        # Write tar entries to a temp .tar file, capture the byte offset
+        # where tarfile will write the EOF marker, then gzip-compress only
+        # the entries (no EOF block).  Omitting the EOF block lets these
+        # per-package gzip streams be raw-concatenated without causing tar
+        # to stop at the first end-of-archive marker.
+        with tempfile.NamedTemporaryFile(suffix=".tar", delete=False, dir=install_tmp) as _tmp_f:
+            tmp_tar_path = _tmp_f.name
```
merschformann: Why do we need tempfile here? Just for the randomized name? Maybe we can use a simple func that gives us a string instead? I am a bit concerned that the file handle is still open when trying to `tarfile.open(tmp_tar_path, mode="w")` right below, particularly on Windows.

nextmv/nextmv/cloud/package.py
```diff
@@ -212,110 +430,778 @@ def __handle_python(
...
+
+        tar_obj = tarfile.open(tmp_tar_path, mode="w")
+        tar_obj.add(install_dir, arcname=dep_arcname)
+        entry_end = tar_obj.offset  # byte offset where EOF marker begins
```
merschformann: Do we need to flush the file, or does this work in all cases? I am just wondering whether buffering might cause us issues here.

nextmv/nextmv/cloud/package.py
```diff
@@ -212,110 +430,778 @@ def __handle_python(
...
+        install_dir = os.path.join(install_tmp, "pkg")
+        os.makedirs(install_dir)
+
+        result = subprocess.run(
+            [
+                uv_bin,
+                "pip",
+                "install",
+                f"{name}=={version}",
+                "--no-deps",
```
merschformann, line 825: > `--no-deps`

Image

merschformann (approved):
I gave it a proper read, but not sure I caught everything. We need some deeper testing across specific cases (deps combinations, platforms, single binary deployments, etc.) before releasing.

However, great work. This will be awesome to have! 🤗

nextmv/nextmv/cli/cache/delete.py
sebastian-quintero: Ich bin dein Freund.

nextmv/nextmv/cache.py
sebastian-quintero: Yeah good point.

nextmv/nextmv/cache.py
sebastian-quintero: Done

nextmv/nextmv/cache.py
sebastian-quintero: You know, I added the `return 0, 0` after the rest of the code was already there, so I kinda forgot to re-read the entire function. You're right, it is stupid.

nextmv/nextmv/cache.py
sebastian-quintero: Here is the AI's reasoning which I agree with:

> The current placement is intentional for one critical reason: rename() is only atomic when source and destination are on the same filesystem. If you stage in tmp and .nextmv lives on a different filesystem (network home directory, separate partition, etc.), Path.rename() raises EXDEV (cross-device link) and you lose atomicity entirely — forcing a copy+delete, which is exactly the partial-write hazard you're trying to avoid.

> On macOS, [tmp](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) is a symlink to [tmp](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) which lives on the root APFS volume, while ~ is typically on a separate APFS Data volume — different filesystems. On Linux, tmp is commonly a tmpfs mount while ~ is on ext4 or xfs. So staging in tmp and renaming into ~/.nextmv would cross a filesystem boundary in both common cases, making rename() fail with EXDEV.

What I did is add a cleanup step in `_create_cache` to remove leftover stuff.

nextmv/nextmv/cache.py
sebastian-quintero: Following the thread above, I am avoiding `tmp` dirs by adding:

```python
if not entry_dir.is_dir() or "-tmp-" in entry_dir.name:
            continue
```

nextmv/nextmv/cache.py
sebastian-quintero: Following the same reasoning:
> Exactly the same reason as store_dep — os.replace() is only atomic when source and destination are on the same filesystem. Using dir=path.parent ensures the temp file is always a sibling of the target, so the final os.replace() is a same-filesystem rename (an inode pointer swap) rather than a copy+delete.

nextmv/nextmv/cloud/package.py
sebastian-quintero: Yeah good catch. Changed to `OSError`.

sebastian-quintero: Must have been copy-paste error.

nextmv/nextmv/cloud/package.py
sebastian-quintero: Hmmm yeah good catch. I was overdoing the whole tempfile a bit too much.

nextmv/nextmv/cloud/package.py
sebastian-quintero: Changed it to use a `with` context manager.

nextmv/nextmv/cloud/package.py
sebastian-quintero: Screw you Python, I know more than you.

sebastian-quintero:
@merschformann addressed all your comments, left some convos open because I don't know if you are happy with my responses. On another note, I tested on 🪟 with @cowanwalls and everything seemed to work without hiccups.

merschformann:
> @merschformann addressed all your comments, left some convos open because I don't know if you are happy with my responses. On another note, I tested on 🪟 with @cowanwalls and everything seemed to work without hiccups.

Can we do a pre-release and run with the .zip deployment too? Or did you already cover that? cc @sebastian-quintero

nextmv/nextmv/cache.py
merschformann: Yeah, that is the reasoning I was expecting. However, the current approach will likely litter the cache directory. Zipping / taring operations and the others done here take a bit of time, a user deciding otherwise may easily kill the CLI in the exact moment (thinking it hangs or something; not necessarily on this one but there are several like this). I would prefer that we at least introduce our own tmp dir if we are so worried about concurrent push commands:

```python
_TMP_DIR = Path.home() / ".nextmv" / "tmp"
```

And just put stuff there under random names. At least, it will be obvious which ones to delete (which we could do or the user). Might even be simple to combine cache cleanup with tmp cleanup (sounds safe to me as the operation shouldn't be run in parallel with something else anyway).

nextmv/nextmv/cache.py
merschformann: Yeah, see the discussion on the first thread. It's what I expected, but I would still prefer for it to be a bit better organized / separated.

merschformann (approved):
If possible I would like to put tmp files that we want to be "atomic" in moving in the `.nextmv/tmp/` folder to have them a bit more isolated for easier cleanup, etc. ... But I wouldn't block on it. I get the motivation for doing it, I am just expecting some litter in folders where I wouldn't want it. See the comments too. ☺️ 

Nothing to block on necessarily though.

Again, thanks for the work. 🤗

sebastian-quintero:
@merschformann I couldn't get the `tmp` to work in a straight forward way. I struggled with cleanup and stuff. I think the way it is right now should be sufficiently stable, and we can figure stuff out should it break for anyone. Leaving it as is, sorry. I did try.

PR #317, opened 2026-05-23
Adds PKCE auth support and login command

merschformann:
# Description

Adds PKCE-based auth (no API key required), team scoping for multi-team accounts, named auth sessions, and system certificate support. Restructures config handling to share logic between `nextmv.cloud` and `nextmv.cli`.

## Changes

- Added `nextmv/auth.py` - full PKCE OAuth2 flow with token storage, OAuth2 `state` parameter (login CSRF protection), token response validation per RFC 6749, `invalid_grant` cleanup on refresh failure, `http://` rejection, `fetch_organizations` for team lookup, and `apply_system_certs` for enterprise proxy support.
- Added `nextmv/config.py` - shared config/sessions I/O, YAML key constants, `AuthType(str, Enum)` for `api_key`/`pkce`, and profile helpers (`get_auth_type`, `get_auth_session`, `get_team_id`, `get_system_certs`, `list_pkce_profiles`, etc.).
- Added `nextmv auth login` and `nextmv auth logout` CLI commands (`cli/auth/`). Login runs the PKCE browser flow per session (deduplicated), supports `--force` for re-authentication via logout-chain. Logout deletes stored tokens per session.
- Extended `nextmv configuration create` to prompt for auth type (or infer `pkce` from OIDC flags), support `--auth-session` for shared sessions, `--team` for team selection, `--system-certs` for enterprise environments, and run the PKCE browser flow inline (reusing valid tokens, skipping the browser when possible). OIDC config stored per-endpoint in `~/.nextmv/sessions.yaml`.
- Extended `cloud/client.py` to detect `pkce` profiles, load/refresh tokens with endpoint-specific OIDC config, set `nextmv-account` header for team scoping, and resolve `system_certs` from constructor → `NEXTMV_SYSTEM_CERTS` env var → config. Defer to explicit `api_key`/`NEXTMV_API_KEY` when provided.
- Added _Type_ and _Session_ columns to `nextmv configuration list`.
- Restructured `cli/main.py` config imports to use the shared `nextmv/config.py`.

## Usage

```bash
# 1. Create a pkce profile (opens browser, picks team)
nextmv configuration create --auth-type pkce --profile work

# 2. Log in (reuses existing token if still valid, otherwise opens browser)
nextmv auth login --profile work

# 3. Use it - any command now picks up the pkce token
nextmv cloud app list --profile work

# 4. (later) Log out of that profile
nextmv auth logout --profile work

# Share a login session across multiple profiles
nextmv configuration create --auth-type pkce --profile dev  --auth-session my-id
nextmv configuration create --auth-type pkce --profile staging --auth-session my-id
nextmv auth login  # logs in both profiles in one browser session

# Enterprise proxy support
nextmv configuration create --auth-type pkce --system-certs
```

Resolves ENG-7402


nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,506 @@
...
+
+# Cognito hosted UI domain for this user pool.
+_COGNITO_DOMAIN = "https://marius-local-nextmv-us-west-2.auth.us-west-2.amazoncognito.com"
+
+# Hard-coded fallback endpoints derived from the discovery document so that
+# the flow works even when the discovery URL is temporarily unreachable.
+_FALLBACK_AUTH_ENDPOINT = f"{_COGNITO_DOMAIN}/oauth2/authorize"
+_FALLBACK_TOKEN_ENDPOINT = f"{_COGNITO_DOMAIN}/oauth2/token"
+
+CLIENT_ID = "dpps8t04d41ns0ua5c59uge0q"
```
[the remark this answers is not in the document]
merschformann: Yep, that is why there is the `TODO`. :relaxed: 

Fair point for making it configurable though. We'll discuss it. :relaxed:

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,506 @@
...
+        The profile name.  ``None`` (or the string ``"default"``) maps to the ``default``
+        sub-directory under ``AUTH_DIR``.
+
+    Returns
+    -------
+    Path
+        The directory path ``~/.nextmv/auth/<name>/``.
+    """
+    name = _DEFAULT_DIR_NAME if (profile is None or profile.strip().lower() == "default") else profile.strip()
+    return AUTH_DIR / name
```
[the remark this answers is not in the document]
merschformann: Fair point, addressed it. :hugs:

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,506 @@
...
+        The profile name.
+    tokens : dict[str, Any]
+        The token dict to persist.  Must contain at least ``access_token``.
+    """
+    path = _token_path(profile)
+    path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
+    with path.open("w") as fh:
+        json.dump(tokens, fh, indent=2)
+    # Restrict read access to the owner only.
+    path.chmod(0o600)
```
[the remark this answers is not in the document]
merschformann: Fair point, will add it.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,506 @@
...
+        If the user does not complete the browser flow within the timeout.
+    RuntimeError
+        If the provider returns an error.
+    requests.HTTPError
+        If the token exchange request fails.
+    """
+    auth_endpoint, token_endpoint = _discover_endpoints()
+    code_verifier, code_challenge = _generate_pkce_pair()
+    redirect_uri = "http://127.0.0.1:56734/callback"
+
```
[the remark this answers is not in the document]
merschformann: Yeah, had to fix the port. Will update the comment. :relaxed:

nextmv/nextmv/cloud/client.py
```diff
@@ -573,6 +585,82 @@ def upload_to_presigned_url(
...
+        # Prefer the id_token when present — API Gateway Cognito authorizers
+        # validate the id_token (which carries the `aud` claim).  Fall back to
+        # access_token for authorizers that accept either.
+        token = tokens.get("id_token") or tokens.get("access_token")
+        if not token:
+            raise ValueError(
+                f"Stored tokens for profile [magenta]{display}[/magenta] do not contain an access token. "
+                f"Please run [code]nextmv login{'  --profile ' + display if profile else ''}[/code] "
+                "to re-authenticate."
+            )
```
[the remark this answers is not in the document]
merschformann: Good point, will simplify since detection of invocation would be quite messy.

nextmv/nextmv/cli/configuration/config.py
```diff
@@ -1,79 +1,29 @@
...
-import yaml
 
 from nextmv import cloud, local
 from nextmv.cli.message import confirmation, error, info, success, warning
 from nextmv.cloud.account import Account
 from nextmv.cloud.client import Client
 from nextmv.cloud.marketplace import MarketplaceApplication, MarketplaceSubscription
 from nextmv.cloud.sso import SSOConfiguration
+from nextmv.config import (
+    CONFIG_DIR,
```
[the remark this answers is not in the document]
merschformann: Yeah, that comment was a left over from the refactor that happened mid way. Let me update. Breaking change should be fine though, since this was CLI only functionality.

nextmv/nextmv/cloud/client.py
```diff
@@ -573,6 +585,82 @@ def upload_to_presigned_url(
...
+        """
+
+        config = _load_config()
+        if not config:
+            return None
+
+        ptype = get_profile_type(config, profile)
+        if ptype != PROFILE_TYPE_AUTH_FLOW:
+            return None
+
```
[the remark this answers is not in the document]
merschformann: Fair point, added some coverage. :+1:

nextmv/nextmv/cloud/client.py
```diff
@@ -573,6 +585,77 @@ def upload_to_presigned_url(
...
+        if is_token_expired(tokens):
+            refresh_token = tokens.get("refresh_token")
+            if not refresh_token:
+                raise ValueError(
+                    f"Access token for profile '{display}' has expired and no refresh token is available. "
+                    f"Please run '{login_cmd}' to re-authenticate."
+                )
+            try:
+                tokens = refresh_tokens(refresh_token)
+                save_tokens(session, tokens)
```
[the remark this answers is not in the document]
merschformann: Good find, thanks! Addressed it. :heart:

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,129 @@
...
+
+    failed: list[str] = []
+
+    for prof in profiles_to_login:
+        display_name = prof if prof is not None else "default"
+        session = get_auth_session(config, prof)
+        endpoint = get_profile_endpoint(config, prof)
+        oidc_cfg = get_endpoint_oidc_config(endpoint, sessions)
+        oidc_discovery_url = oidc_cfg.get(OIDC_DISCOVERY_URL_KEY) if oidc_cfg else None
+        oidc_client_id = oidc_cfg.get(CLIENT_ID_KEY) if oidc_cfg else None
```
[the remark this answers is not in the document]
merschformann: Thanks, addressed it with a skip and a warning.

nextmv/nextmv/config.py
```diff
@@ -0,0 +1,393 @@
...
+        set.
+    """
+    if profile is None:
+        ep = config.get(ENDPOINT_KEY, DEFAULT_ENDPOINT)
+    else:
+        profile_data = config.get(profile, {})
+        ep = profile_data.get(ENDPOINT_KEY, DEFAULT_ENDPOINT) if isinstance(profile_data, dict) else DEFAULT_ENDPOINT
+
+    ep = str(ep)
+    return _strip_scheme(ep) or DEFAULT_ENDPOINT
```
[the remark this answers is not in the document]
merschformann: Thanks, improved it.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,604 @@
...
+    redirect_uri = f"http://127.0.0.1:{CALLBACK_PORT}"
+
+    params = {
+        "response_type": "code",
+        "client_id": cid,
+        "redirect_uri": redirect_uri,
+        "scope": SCOPES,
+        "code_challenge_method": "S256",
+        "code_challenge": code_challenge,
+    }
```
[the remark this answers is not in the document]
merschformann: Good point, improved it.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,604 @@
...
+
+    listener.join(timeout=_BROWSER_TIMEOUT + 5)
+
+    if exc_holder:
+        raise exc_holder[0]
+
+    code = callback_result.get("code")
+    if not code:
+        raise RuntimeError("No authorization code received. Please try running [code]nextmv login[/code] again.")
+
```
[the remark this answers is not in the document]
merschformann: Yep, addressed it.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,604 @@
...
+    # Handle exactly one request.
+    server.handle_request()
+    server.server_close()
+
+    params: dict[str, str] = server.callback_params  # type: ignore[attr-defined]
+    if not params:
+        raise TimeoutError(
+            f"No callback received within {_BROWSER_TIMEOUT} seconds. "
+            "Please try running [code]nextmv login[/code] again."
+        )
```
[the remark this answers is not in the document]
merschformann: Yep, addressed it.

nextmv/nextmv/cli/configuration/list.py
```diff
@@ -45,15 +47,17 @@ def list() -> None:
             "name": k,
             "api_key": v.get(API_KEY_KEY),
             "endpoint": v.get(ENDPOINT_KEY),
+            "profile_type": v.get(PROFILE_TYPE_KEY, "api_key"),
```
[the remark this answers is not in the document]
merschformann: Sure, why not.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+            "--auth-type",
+            "-t",
+            help=(
+                "The authentication type for this profile: [magenta]api_key[/magenta] (default) or "
+                "[magenta]pkce[/magenta] (browser-based PKCE login via [code]nextmv login[/code]). "
+                "Ignored when [magenta]--api-key[/magenta] is provided."
+            ),
+            metavar="AUTH_TYPE",
+        ),
+    ] = None,
```
sebastian-quintero, line 88:
```suggestion
    ] = "api_key",
```

Setting as default allows you to omit the `(default)` mention in the help message.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
     ] = None,
+    auth_type: Annotated[
+        str | None,
+        typer.Option(
+            "--auth-type",
+            "-t",
+            help=(
+                "The authentication type for this profile: [magenta]api_key[/magenta] (default) or "
+                "[magenta]pkce[/magenta] (browser-based PKCE login via [code]nextmv login[/code]). "
+                "Ignored when [magenta]--api-key[/magenta] is provided."
```
sebastian-quintero:
```suggestion
                "Ignored when --api-key is provided."
```

Typer should auto-color the option. When talking about options we don't use the magenta.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
         ),
     ] = None,
+    auth_type: Annotated[
+        str | None,
+        typer.Option(
+            "--auth-type",
+            "-t",
+            help=(
+                "The authentication type for this profile: [magenta]api_key[/magenta] (default) or "
+                "[magenta]pkce[/magenta] (browser-based PKCE login via [code]nextmv login[/code]). "
```
sebastian-quintero: I would move the `(browser-based PKCE login via [code]nextmv login[/code])` to the command's help, to keep the option's help from becoming too long.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+        typer.Option(
+            "--auth-session",
+            "-s",
+            help=(
+                "Named auth session to share tokens across profiles. "
+                "Only applies to [magenta]pkce[/magenta] profiles. "
+                "Multiple profiles that reference the same session name share a single "
+                "browser login. "
+                f"Defaults to the reserved [magenta]{DEFAULT_AUTH_SESSION}[/magenta] session "
+                "when omitted."
```
sebastian-quintero: I would move most oh this help text to the main command's help, to about the option itself from becoming too long.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+            ),
+            metavar="SESSION_NAME",
+        ),
+    ] = None,
```
sebastian-quintero:
```suggestion
    ] = DEFAULT_AUTH_SESSION,
```

Should this be the default?

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+    team: Annotated[
+        str | None,
+        typer.Option(
+            "--team",
+            help=(
+                "Team name to associate with this [magenta]pkce[/magenta] profile. "
+                "When omitted, available teams are fetched from the API and you will be "
+                "prompted to select one. "
```
sebastian-quintero: This more detailed part of the explanation I would move to the main help.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+        ),
+    ] = None,
+    oidc_discovery_url: Annotated[
+        str | None,
+        typer.Option(
+            "--oidc-discovery-url",
+            hidden=True,
+            help=(
+                "OIDC discovery document URL for the identity provider behind the endpoint. "
+                "Only needed for non-production endpoints not already in sessions.yaml. "
```
sebastian-quintero: Same here, move the longer part of the explanation to the main help of the command.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+                "Only applies to [magenta]pkce[/magenta] profiles."
+            ),
+            metavar="TEAM_NAME",
+        ),
+    ] = None,
+    oidc_discovery_url: Annotated[
+        str | None,
+        typer.Option(
+            "--oidc-discovery-url",
+            hidden=True,
```
sebastian-quintero, line 115: Why is this option and the other `oidc` one hidden? Is it internal only?

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+            help=(
+                "OIDC discovery document URL for the identity provider behind the endpoint. "
+                "Only needed for non-production endpoints not already in sessions.yaml. "
+                "Only applies to [magenta]pkce[/magenta] profiles."
+            ),
+            metavar="OIDC_DISCOVERY_URL",
+        ),
+    ] = None,
+    oidc_client_id: Annotated[
```
sebastian-quintero: We use alphabetical ordering for the options. Please change given the options of this command are currently not in alphabetical order.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
-    - Default configuration.
+    - Default configuration (prompts for type and API key or opens browser).
         $ [dim]nextmv configuration create[/dim]
 
-    - Default configuration without prompting for an API key.
+    - Default API key configuration without prompting.
         $ [dim]nextmv configuration create --api-key NEXTMV_API_KEY[/dim]
 
-    - Configure a profile named [magenta]hare[/magenta].
+    - Configure a named [magenta]api_key[/magenta] profile.
```
sebastian-quintero:
```suggestion
    - Configure an [magenta]api_key[/magenta] profile named [magenta]hare[/magenta].
```

Keeping the original text.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
     if profile is not None and profile.strip().lower() == "default":
         error("[magenta]default[/magenta] is a reserved profile name.")
 
-    endpoint = str(endpoint)
-    if endpoint.startswith("https://"):
-        endpoint = endpoint[len("https://") :]
-    elif endpoint.startswith("http://"):
-        endpoint = endpoint[len("http://") :]
+    # Validate and normalise auth_session.
+    if auth_session is not None:
```
sebastian-quintero, line 183: You can break this function up now. Move auth_session logic to a sub-functoin, and the api key stuff to another sub function. That way we don't have to use the `noqa C901` exclusion.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+    # If --api-key is supplied, we always use the api_key auth type regardless
+    # of --auth-type, since there's an explicit credential.
+    if api_key is not None and api_key.strip():
+        resolved_type = AUTH_TYPE_API_KEY
+    elif auth_type is not None:
+        auth_type = auth_type.strip().lower()
+        if auth_type not in (AUTH_TYPE_API_KEY, AUTH_TYPE_PKCE):
+            error(
+                f"Invalid auth type [magenta]{auth_type}[/magenta]. "
+                f"Must be [magenta]{AUTH_TYPE_API_KEY}[/magenta] or [magenta]{AUTH_TYPE_PKCE}[/magenta]."
```
sebastian-quintero: We can overcome this by using a `class AuthType(str, Enum)` that enumerates the types. Typer would already validate the enumerations. Take a look at [`create.py`](https://github.com/nextmv-io/nextmv-py/blob/develop/nextmv/nextmv/cli/cloud/run/create.py): `content_format` and `run_type` options show how to use an enum class for these purposes.

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+            if profile not in config:
+                config[profile] = {}
+            config[profile][AUTH_TYPE_KEY] = AUTH_TYPE_PKCE
+            config[profile][ENDPOINT_KEY] = endpoint
+            config[profile][TEAM_ID_KEY] = team_id
+            config[profile].pop(API_KEY_KEY, None)
+            if auth_session:
+                config[profile][AUTH_SESSION_KEY] = auth_session
+            else:
+                config[profile].pop(AUTH_SESSION_KEY, None)
```
sebastian-quintero: This got me wondering: is there a better place where we can handle profiles? Like writing to and reading from them? Maybe `config.py`. When using PKCE or the API key this all seems very manual for the CLI. Not a big deal though.

nextmv/nextmv/cli/configuration/list.py
```diff
@@ -32,6 +41,8 @@ def list() -> None:
     default = {
         "api_key": config.get(API_KEY_KEY),
         "endpoint": config.get(ENDPOINT_KEY),
+        "auth_type": config.get(AUTH_TYPE_KEY, "api_key"),
```
sebastian-quintero, line 45: To my point about having a class for enumerating `AuthType`, it would prevent declaring verbatim strings that we need to maintain.

nextmv/nextmv/cli/configuration/list.py
```diff
@@ -40,20 +51,30 @@ def list() -> None:
...
             "endpoint": v.get(ENDPOINT_KEY),
+            "auth_type": v.get(AUTH_TYPE_KEY, "api_key"),
+            "auth_session": get_auth_session(config, k) if v.get(AUTH_TYPE_KEY) == AUTH_TYPE_PKCE else None,
         }
         profiles.append(profile)
 
-    table = Table("Profile name", "API Key", "Endpoint")
+    table = Table("Profile name", "Type", "Session", "API Key", "Endpoint")
     not_set = "[italic]Not set[/italic]"
+    n_a = "[italic dim]—[/italic dim]"
```
sebastian-quintero: LOL why not `na`?

nextmv/nextmv/cli/configuration/list.py
```diff
@@ -69,6 +90,8 @@ def list() -> None:
 
         table.add_row(
             f"[bold yellow]{profile['name']}[/bold yellow]",
+            f"[bold yellow]{profile['auth_type']}[/bold yellow]",
+            f"[bold yellow]{session_display}[/bold yellow]",
```
sebastian-quintero, line 95: I wonder if we should build the table conditionally? Or even have two tables that are printed one after the other (one for API keys the other for sessions)? This is to avoid empty columns for the normal users that only have API keys... Not a biggie though, I can live with this.

nextmv/nextmv/cli/auth/login.py
```diff
@@ -0,0 +1,171 @@
...
+    load_config,
+    load_sessions,
+)
+
+# Set up subcommand application.
+app = typer.Typer(invoke_without_command=True)
+
+
+@app.callback()
+def login(
```
sebastian-quintero, line 37: Should this rather be `nextmv configuration login`? If possible, I would like to prevent adding many commands to the root. However, `nextmv configuration login` might read as "the configuration is logging in" which is absurd. Would there be a good case for nesting it under configuration and that way we can keep all commands related to profiles and CLI auth/config nested under `configuration`?

nextmv/nextmv/cli/auth/login.py
```diff
@@ -0,0 +1,171 @@
...
+        str | None,
+        typer.Option(
+            "--profile",
+            "-p",
+            help="Profile to log in to. If omitted, all [magenta]pkce[/magenta] profiles are logged in.",
+            envvar="NEXTMV_PROFILE",
+            metavar="PROFILE_NAME",
+        ),
+    ] = None,
+    force: Annotated[
```
sebastian-quintero, line 38: Alphabetical ordering of options in commands

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,171 @@
...
+        ),
+    ] = False,
+) -> None:
+    """
+    Log in to Nextmv using the browser-based PKCE auth flow.
+
+    Opens your browser, completes the OAuth2 PKCE flow, and stores the
+    resulting tokens under [magenta]~/.nextmv/auth/[/magenta].
+
+    If [magenta]--profile[/magenta] is given, only that profile is logged in
```
sebastian-quintero:
```suggestion
    If --profile is given, only that profile is logged in
```

Don't use magenta style for command options, as those are colored automatically.

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,171 @@
...
+    (it must be configured with [code]auth_type: pkce[/code]).  If no
```
sebastian-quintero:
```suggestion
    (it must be configured with [code]--auth_type[/code] set to [magenta]pkce[/magenta] with command [code]nextmv configuration create[/code]). If no
```

More explicit is always better.

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,171 @@
...
+        profiles_to_login: list[str | None] = [profile]
+    else:
+        profiles_to_login = list_pkce_profiles(config)
+
+    if not profiles_to_login:
+        info(
+            "No [magenta]pkce[/magenta] profiles found. "
+            "Use [code]nextmv configuration create[/code] to create one, "
+            "or set [magenta]auth_type: pkce[/magenta] in "
+            "[magenta]~/.nextmv/config.yaml[/magenta]."
```
sebastian-quintero: I would prefer we don't tell people to touch the config files by hand.

nextmv/nextmv/cli/auth/login.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+def _run_login(profiles_to_login: list[str], config: dict, sessions: dict, force: bool) -> None:
+    """Run the login flow for the given list of profiles."""
+    failed: list[str] = []
+    # Deduplicate: one browser flow per unique (session, endpoint) pair.
+    # Multiple profiles can share the same auth session — logging in once is enough
+    # because tokens are stored per session, not per profile.
+    seen_sessions: dict[tuple[str, str], list[str]] = {}  # (session, endpoint) -> [display_name, ...]
+    session_oidc: dict[tuple[str, str], tuple[str | None, str | None]] = {}
```
sebastian-quintero, line 130: The extremely typed Python can get annoying LOL. I think defining vars without a type hint is fine, if that is your preference.

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,171 @@
...
+            failed.append(display_name)
+            continue
+        key = (session, endpoint)
+        seen_sessions.setdefault(key, []).append(display_name)
+        session_oidc[key] = (oidc_cfg.get(OIDC_DISCOVERY_URL_KEY), oidc_cfg.get(CLIENT_ID_KEY))
+
+    for (session, endpoint), profile_names in seen_sessions.items():
+        oidc_discovery_url, oidc_client_id = session_oidc[(session, endpoint)]
+        profiles_display = ", ".join(f"[magenta]{n}[/magenta]" for n in profile_names)
+        message(
```
sebastian-quintero: Should this use `in_progress` instead of `message`?

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+
+import requests
+
+AUTH_DIR = Path.home() / ".nextmv" / "auth"
+
+# >>> Provider constants
+
+OIDC_DISCOVERY_URL = "https://cognito-idp.us-east-2.amazonaws.com/us-east-2_1jHS2b9HU/.well-known/openid-configuration"
+CLIENT_ID = "4k91vdlr1m52v9v45h9vc7a25e"
+SCOPES = "email openid profile"
```
sebastian-quintero, line 74: Is all of this ok to be made public?

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    Returns
+    -------
+    tuple[str, str, str]
+        ``(authorization_endpoint, token_endpoint, logout_endpoint)``
+    """
+    discovery_url = oidc_discovery_url or OIDC_DISCOVERY_URL
+    try:
+        resp = requests.get(discovery_url, timeout=10)
+        resp.raise_for_status()
+        doc = resp.json()
```
sebastian-quintero, line 396: Maybe we should use our own `Client.requests` with empty information to perform requests. It already handles all the good stuff like timeout, exponential backoff with jitter, checking status, etc.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    tuple[str, str]
+        ``(code_verifier, code_challenge)``
+    """
+    code_verifier = secrets.token_urlsafe(64)
+    digest = hashlib.sha256(code_verifier.encode("ascii")).digest()
+    code_challenge = base64.urlsafe_b64encode(digest).rstrip(b"=").decode("ascii")
+    return code_verifier, code_challenge
+
+
+class _CallbackHandler(http.server.BaseHTTPRequestHandler):
```
sebastian-quintero, line 431: Oh God do we really need this class?

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    """
+    cid = client_id or CLIENT_ID
+    payload = {
+        "grant_type": "authorization_code",
+        "client_id": cid,
+        "code": code,
+        "redirect_uri": redirect_uri,
+        "code_verifier": code_verifier,
+    }
+    resp = requests.post(
```
sebastian-quintero, line 625: Same question about using our own `Client.request`.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        The list of organization objects returned by the API.
+
+    Raises
+    ------
+    requests.HTTPError
+        If the API returns a non-2xx response.
+    """
+    # Ensure the endpoint has a scheme.
+    base = endpoint if endpoint.startswith(("https://", "http://")) else f"https://{endpoint}"
+    url = f"{base.rstrip('/')}/v1/internal/me/organization"
```
sebastian-quintero: Are we ok using/exposing `internal` endpoints??

nextmv/nextmv/config.py
```diff
@@ -0,0 +1,393 @@
...
+AUTH_SESSION_KEY = "auth_session"
+TEAM_ID_KEY = "team_id"
+
+# Sessions keys
+OIDC_DISCOVERY_URL_KEY = "oidc_discovery_url"
+CLIENT_ID_KEY = "client_id"
+
+# Auth type values
+AUTH_TYPE_API_KEY = "api_key"
+AUTH_TYPE_PKCE = "pkce"
```
sebastian-quintero: Yep, seems like a good idea to place in an enum class.

sebastian-quintero (changes requested):
I'm only marking as blocked due to the question around secrets and `internal`. Otherwise would have marked as `Comment`. This looks like quite the work! Quite involved as well. Did my best to read through the whole thing.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    ----------
+    session : str
+        The auth session name.  Resolve this from a profile via
+        :func:`nextmv.config.get_auth_session`.
+    tokens : dict[str, Any]
+        The token dict to persist.  Must contain at least ``access_token``.
+    """
+    path = _token_path(session)
+    path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
+    with path.open("w") as fh:
```
dirkschumacher: Is there a way to apply the mode directly when opening the file? e.g. something like that?
```
os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
```

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    list[dict[str, Any]]
+        The list of organization objects returned by the API.
+
+    Raises
+    ------
+    requests.HTTPError
+        If the API returns a non-2xx response.
+    """
+    # Ensure the endpoint has a scheme.
+    base = endpoint if endpoint.startswith(("https://", "http://")) else f"https://{endpoint}"
```
dirkschumacher: I think we should reject `http://` here

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        The OIDC discovery document URL to use.  When ``None`` the module-level
+        constant :data:`OIDC_DISCOVERY_URL` is used (which covers the production
+        endpoint).
+
+    Returns
+    -------
+    tuple[str, str, str]
+        ``(authorization_endpoint, token_endpoint, logout_endpoint)``
+    """
+    discovery_url = oidc_discovery_url or OIDC_DISCOVERY_URL
```
dirkschumacher, line 392: Should we really fallback if the user gives an explicit discovery url here? It seems better to fail IMHO.

nextmv/nextmv/config.py
```diff
@@ -0,0 +1,393 @@
...
+    if not CONFIG_FILE.exists():
+        return {}
+
+    with CONFIG_FILE.open() as file:
+        config = yaml.safe_load(file)
+
+    return config if isinstance(config, dict) else {}
+
+
+def save_config(config: dict[str, Any]) -> None:
```
dirkschumacher, line 149: Should we also apply chmod here in a best effort way similar to the tokens file?

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    ----------
+    oidc_discovery_url : str | None
+        The OIDC discovery document URL for the identity provider backing this
+        profile's endpoint.  When ``None`` the module-level
+        :data:`OIDC_DISCOVERY_URL` constant is used (production endpoint).
+    client_id : str | None
+        The OAuth2 client ID for the identity provider.  When ``None`` the
+        module-level :data:`CLIENT_ID` constant is used.
+    force : bool
+        When ``True``, opens the Cognito logout endpoint first (``/logout``),
```
dirkschumacher: Is this docstring part of the public api? If so I think we should not leak cognito as a concept here into the public api. This can change in the future. Maybe rephrasing this docstring does solve it.

nextmv/nextmv/cli/login.py
```diff
@@ -0,0 +1,171 @@
...
+            "or set [magenta]auth_type: pkce[/magenta] in "
+            "[magenta]~/.nextmv/config.yaml[/magenta]."
+        )
+        return
+
+    # Run the login flow for the selected profiles.
+    _run_login(profiles_to_login, config, sessions, force)
+
+
+def _run_login(profiles_to_login: list[str], config: dict, sessions: dict, force: bool) -> None:
```
dirkschumacher: `list_pkce_profiles` returns `list[str | None]` but `profiles_to_login` has type annotation `list[str]`

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -7,29 +7,43 @@
...
+    AUTH_TYPE_API_KEY,
+    AUTH_TYPE_KEY,
+    AUTH_TYPE_PKCE,
+    CLIENT_ID_KEY,
+    DEFAULT_AUTH_SESSION,
     DEFAULT_ENDPOINT,
     ENDPOINT_KEY,
+    OIDC_DISCOVERY_URL_KEY,
+    TEAM_ID_KEY,
+    _strip_scheme,
```
dirkschumacher, line 39: Is the underscore a convention that the function is private? Then maybe it should be public

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        The query parameters extracted from the callback URL.
+
+    Raises
+    ------
+    TimeoutError
+        If no callback is received within ``_BROWSER_TIMEOUT`` seconds.
+    RuntimeError
+        If the provider returns an error parameter.
+    """
+    server = http.server.HTTPServer(("127.0.0.1", port), _CallbackHandler)
```
dirkschumacher: Does this fail if the port is already taken? Maybe we can give the user a friendly error message by wrapping that into try catch

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+
+class _CallbackHandler(http.server.BaseHTTPRequestHandler):
+    """
+    Minimal HTTP handler that captures the OAuth2 callback query parameters.
+
+    The captured parameters are stored in ``self.server.callback_params``.
+    """
+
+    def do_GET(self) -> None:  # noqa: N802
+        parsed = urllib.parse.urlparse(self.path)
```
dirkschumacher, line 472: To make it more robust we could return a 404 unless `self.path == "/"`.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    try:
+        expires_at = datetime.fromisoformat(expires_at_str)
+        # Ensure timezone-aware comparison.
+        if expires_at.tzinfo is None:
+            expires_at = expires_at.replace(tzinfo=timezone.utc)
+        now = datetime.now(tz=timezone.utc)
+        # Consider expired if within 30-second buffer.
+        return (expires_at - now).total_seconds() < 30
+    except ValueError:
+        # Unparseable expiry — treat as not expired.
```
dirkschumacher: Why do we treat it is as not expired? Wouldn't it be safer to do the opposite? I think the only case where is happens if the token is corrupt.

nextmv/nextmv/config.py
```diff
@@ -0,0 +1,393 @@
...
+# Sessions keys
+OIDC_DISCOVERY_URL_KEY = "oidc_discovery_url"
+CLIENT_ID_KEY = "client_id"
+
+# Auth type values
+AUTH_TYPE_API_KEY = "api_key"
+AUTH_TYPE_PKCE = "pkce"
+
+# Defaults
+DEFAULT_AUTH_SESSION = "default"
```
dirkschumacher: `_DEFAULT_SESSION_NAME` also defines default. Maybe use `DEFAULT_AUTH_SESSION` instead of `_DEFAULT_SESSION_NAME`?

nextmv/nextmv/cli/configuration/create.py
```diff
@@ -54,60 +68,407 @@ def create(
...
+                    refresh_token,
+                    client_id=client_id,
+                    oidc_discovery_url=oidc_discovery_url,
+                )
+                save_tokens(session, tokens)
+                token = tokens.get("id_token") or tokens.get("access_token")
+                if token:
+                    return token
+            except Exception:
+                pass  # Fall through to full browser flow.
```
dirkschumacher: Should we send a message to the user that this failed?

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        return auth_ep, token_ep, logout_ep
+    except Exception:
+        return _FALLBACK_AUTH_ENDPOINT, _FALLBACK_TOKEN_ENDPOINT, _FALLBACK_LOGOUT_ENDPOINT
+
+
+def _generate_pkce_pair() -> tuple[str, str]:
+    """
+    Generate a PKCE ``code_verifier`` and its ``code_challenge``.
+
+    The verifier is a cryptographically random URL-safe string (43-128 chars as per
```
dirkschumacher: Can we add a reference to the RFC here? https://datatracker.ietf.org/doc/html/rfc7636#section-4.2

This implements S256 which is great.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        return (expires_at - now).total_seconds() < 30
+    except ValueError:
+        # Unparseable expiry — treat as not expired.
+        return False
+
+
+# >>> PKCE flow — internal helpers
+
+
+def _discover_endpoints(oidc_discovery_url: str | None = None) -> tuple[str, str, str]:
```
dirkschumacher, line 364: We should reject urls with `http://`. This is inline with the RFCs

nextmv/nextmv/auth.py
dirkschumacher: The RFC wants 43 to 128 chars. But this function always produces a fixed length of 86 chars.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+        parsed = urllib.parse.urlparse(self.path)
+        params = dict(urllib.parse.parse_qsl(parsed.query))
+        # Store on the server instance so the main thread can read them.
+        self.server.callback_params = params  # type: ignore[attr-defined]
+
+        if "error" in params:
+            body = (
+                "<html><body>"
+                "<h2>Authentication failed.</h2>"
+                f"<p>{params.get('error_description', params['error'])}</p>"
```
dirkschumacher: We should html escape that

nextmv/nextmv/cloud/client.py
```diff
@@ -573,6 +592,99 @@ def upload_to_presigned_url(
...
+                oidc_cfg = get_endpoint_oidc_config(endpoint, load_sessions())
+                tokens = refresh_tokens(
+                    refresh_token,
+                    oidc_discovery_url=oidc_cfg.get(OIDC_DISCOVERY_URL_KEY) if oidc_cfg else None,
+                    client_id=oidc_cfg.get(CLIENT_ID_KEY) if oidc_cfg else None,
+                )
+                save_tokens(session, tokens)
+            except Exception as exc:
+                raise ValueError(
+                    f"Failed to refresh access token for profile '{display}': {exc}. "
```
dirkschumacher, line 937: Maybe we should delete the tokens from disk when refresh fails because of `invalid_grant`.

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+    resp = requests.post(
+        token_endpoint,
+        data=payload,
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+        timeout=30,
+    )
+    try:
+        resp.raise_for_status()
+    except requests.HTTPError as exc:
+        raise requests.HTTPError(f"Token exchange failed ({resp.status_code}): {resp.text}") from exc
```
dirkschumacher, line 634: The error response is always a json. Not sure if we can format that better here. Cosmetic change. Not sure

nextmv/nextmv/auth.py
```diff
@@ -0,0 +1,637 @@
...
+
+    tokens: dict[str, Any] = resp.json()
+    # Compute and store an absolute expiry timestamp.
+    expires_in: int = tokens.get("expires_in", 3600)
+    expires_at = datetime.now(tz=timezone.utc) + timedelta(seconds=expires_in)
+    tokens["expires_at"] = expires_at.isoformat()
+    return tokens
```
dirkschumacher, line 642: Not sure if this is the best place but we have to validate the response (or it is already happening):
https://datatracker.ietf.org/doc/html/rfc6749#section-5.1

the tokens need to have `access_token` and `token_type`. For the latter we only support "bearer" (https://datatracker.ietf.org/doc/html/rfc6749#section-7.1).
expires_in is optional, so that is fine here.

nextmv/nextmv/auth.py
dirkschumacher: Yes, that is fine.

nextmv/nextmv/cli/configuration/create.py
merschformann: The one on `_strip_scheme`? I don't think we want to signal to users that this is intended functionality of the Python SDK, right?

nextmv/nextmv/cli/configuration/create.py
merschformann: Yes, it's meant for us as we work with local deploys and dev/staging environments. Users will for now only interface with prod. However, this would change when self-deploy is publicly supported. ☺️ 
Or, if we add support for custom auth providers.

nextmv/nextmv/cli/configuration/create.py
merschformann: Yeah, I restructured it. I hope it makes sense to you. I like the api-key / pkce split.

nextmv/nextmv/cli/configuration/create.py
merschformann: Not sure, I wouldn't want to unnecessarily pollute the output / confused the user (make them think something is wrong when it may not be), since there is a clear path forward (continuing in the browser). 🤷

nextmv/nextmv/cli/configuration/create.py
merschformann: Yep, using the new enum type here now. Good call.

nextmv/nextmv/cli/configuration/list.py
merschformann: Yeah, that enum is here now. ☺️

nextmv/nextmv/cli/configuration/list.py
merschformann: Yeah, I considered it too. But I think I prefer the simple one table, as I feel two would be more confusing. And the complexity is not huge, even if one has as many profiles as I have by now. 😂 

But yeah, happy to discuss further / vote about it. But if not pressing, I would wait for user feedback around this.

nextmv/nextmv/cli/auth/login.py
merschformann: I personally prefer the root level `login` as it's so commonly used. However, I do get your point. Not sure what the best way to deal with it is. We could think about `nextmv auth login` or similar. That would also allow putting `nextmv auth logout`, which @dirkschumacher was wishing for in it. 

Not sure, what do you all think?

nextmv/nextmv/cli/auth/login.py
merschformann: You are quite strict. 😂 

But yeah, I addressed it. 😊

nextmv/nextmv/cli/auth/login.py
merschformann: I go back and forth on it, I do like that pylance picks up on more things, but it can look weird. As a former old-schoolish OO person, I don't mind the additional type definitions. So, I will leave them for now. But also fine to remove. 🍅  / 🍅  to me. 😅

nextmv/nextmv/cloud/client.py
merschformann: Not sure, is that recommended? I wouldn't want to delete tokens if it was just a blip, no? 🤔

nextmv/nextmv/auth.py
merschformann: Yep, changed it. ❤️ That should be a better experience (either refreshes the token or makes the user login again).

Thanks for the find.

nextmv/nextmv/auth.py
merschformann: Changed it.

nextmv/nextmv/auth.py
merschformann: Yeah, agreed. Changed it. ❤️

nextmv/nextmv/auth.py
merschformann: That would be tricky (circular imports, etc.) and feels like having a separate more simple implementation is nicer for the more security critical auth process. So, the split is kind of intentional.

nextmv/nextmv/auth.py
merschformann: But 64 should result in 86 which is in range [43, 128]. Are you sure that it should be a randomly chosen length? And if so, what distribution should be used? Uniform? 🤔 

Probably won't hurt / might help. Just not how I understood it. 😂

dirkschumacher: No I am saying the docs say this function returns a random string (43-128 chars) but the code returns a fixed length. We should reflect that in the docs.

merschformann: Oh yeah, the doc was basically the excerpt from the RFC. Good point. Will change. :relaxed:

nextmv/nextmv/cloud/client.py
dirkschumacher: `invalid_grant` is never temporary. When the server says the refresh token is invalid then the token won't become valid again.

merschformann: Ah ok, I read it more like if refresh fails in general and not just for the specific reason.

dirkschumacher: No, just for that reason. Of course the server could be down or anything. But if it is `invalid_grant` we know it's invalid.

merschformann: Sure. What is the format that we return that error in? Is it a 400 with this body? `{"error":"invalid_grant"}`

nextmv/nextmv/auth.py
merschformann: We would need to catch a formatting error if it doesn't convert though. Not sure whether that is worth it. :shrug: Feels a bit like overengineering it.

nextmv/nextmv/auth.py
merschformann: Adding it.

nextmv/nextmv/auth.py
merschformann: Yes. It's the minimal way to implement a one-shot HTTP server callback handler using stdlib `http.server`. The alternative would be a raw socket listener, which would be more code and less readable. ☺️

nextmv/nextmv/cloud/client.py
dirkschumacher: Yes, see here: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

merschformann: Yes, that field I read. I just wasn't sure we whether I can fully rely on it. :sweat_smile:

nextmv/nextmv/auth.py
merschformann: 🫡

nextmv/nextmv/auth.py
merschformann: I think I replied to this earlier, but just in case I am misremembering: The split implementation is mostly intentional (besides the potential circular import). For the security critical flow here, I think it is healthier to separate concerns here. 😊

nextmv/nextmv/config.py
merschformann: Yeah, good catch. Too many definitions and their constants changed in this PR over time. 😅

nextmv/nextmv/config.py
merschformann: Yeah, good point. We never did so far, but the `config.yaml` does contain the API keys of course.

nextmv/nextmv/cli/auth/login.py
dirkschumacher: I have no strong feelings here. I have seen something like `cli auth login` or `cli sso login`. But I have also seen `cli login` in popular tools. I can live with both 😅

merschformann: I will change it to `nextmv auth login`. That fits the structure and should be on point (in contrast to `nextmv sso login` for example). Also, I don't want to merge it with config management, as that already grows quite a bit atm. The login feels nicer when separated. However, introducing the additional layer gives the opportunity to have the `logout` too, apart from better adhering with the existing structure. :relaxed:

nextmv/nextmv/auth.py
merschformann: Done.

nextmv/nextmv/cloud/client.py
merschformann: Done.

nextmv/nextmv/auth.py
merschformann: Good call, added it.

nextmv/nextmv/cli/configuration/create.py
merschformann: I hope it's sufficiently short now. :blush: 
Otherwise, let me know. I can go over it again. I might have gotten lost in the flood of comments. :sweat_smile:

nextmv/nextmv/auth.py
cowanwalls, line 1: Not the worst (in terms of reading experience) auth handler file I've read. I think this looks fine but given the size of the file I am not sure I 100% have the level of detail I'd want on the review. The good news is that obvious issues with this should at least get surfaced in testing (and I do intend to do testing)

cowanwalls (commented):
I have no notable comments. I will do some manual testing of this branch tomorrow and re-review (if it is not already merged).

nextmv/nextmv/auth.py
merschformann: Thank you @cowanwalls ! :hugs:

nextmv/nextmv/cli/auth/login.py
sebastian-quintero: ok

sebastian-quintero (approved):
Thank you for all the work.

nextmv/nextmv/cloud/client.py
cowanwalls: Seems like this might be a line we'd need to flag the linter to ignore. Not really sure there is a way that's both good and pythonic to address it in this case.

cowanwalls (approved):
Approving. Can re-approve once the linting error is either fixed or ignored if needed

nextmv/nextmv/cloud/client.py
cowanwalls: Oh silly me I don't know how I ended up on an outdated commit

PR #320, opened 2026-06-01
Fix content_format when submitting ensemble run with CLI

sebastian-quintero:
# Description

There was a usage bug that requires a user to specify the `--content-format` option when submitting an ensemble run with the CLI. Now, we are inferring the content format based on the provided input, so even though in the backend the content format is required, the CLI resolves that for the user.

Also added a bit more validations around ensemble runs now that I was in there.


merschformann (approved):
Nice, looks good. ☺️

PR #321, opened 2026-06-01
Fixes YAML parsing by removing explicit newline characters (\n)

sebastian-quintero:
# Description

When parsing the `app.yaml` manifest using block scalar styles (`>`, `>-`, `|`, `|-`), the newlines some of them introduced were preserved as explicit newline characters `\n`. The API doesn't like this, so it returned a 400 bad request error. This PR addresses this behavior by stripping these newline characters and replacing them with spaces.


merschformann (approved):
Thank you! ❤️

PR #323, opened 2026-06-03
Display nice error messages for HTTP errors in CLI and enable the `--debug` option in every command

sebastian-quintero:
# Description

- Catch HTTP exceptions and display some nice error messages.
- Creates the `--debug` option and adds it to every command. When this option is used, we don't pretty-print errors, but rather dump the entire Python traceback so that it is easy to spot the bug.

### DIFF
--- a/nextmv/nextmv/cli/main.py
+++ b/nextmv/nextmv/cli/main.py
@@ -20,7 +20,7 @@ import sys
 import warnings
 from typing import Annotated
 
-import rich
+import requests
 import typer
 from typer import rich_utils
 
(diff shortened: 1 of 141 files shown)


nextmv/nextmv/cli/main.py
```diff
@@ -309,7 +350,7 @@ def main() -> None:
     """
 
     # Improve compatibility with Windows terminals.
-    setup_encoding()
+    _setup_encoding()
```
merschformann, line 353: ❤️

merschformann (approved):
Nice! Thank you for this! 🤗

PR #324, opened 2026-06-08
Bumps uv dependencies to resolve dependabot issues

merschformann:
# Description

Bumps uv dependencies to address dependabot issues.


cowanwalls (approved):
Well, time to hit the Big Friendly Red Button I suppose.

PR #326, opened 2026-06-10
Enable pagination for listing entities - SDK & CLI

sebastian-quintero:
# Description

Enables pagination _by default_ on all functions that list entities, for both the SDK and CLI.

- Adds `no_pagination: bool` to all `list_xyz` methods and functions under `cloud` in the SDK. By default, pagination is enabled, meaning `no_pagination` has a default value of `False`.
- Adds `--no-pagination` to all `list` commands under `nextmv cloud`. By default, pagination is enabled, meaning that `--no-pagination` would disable pagination.
- Adds integration SDK integration tests to make sure we can list with and without pagination. Adds missing integration tests for managed inputs.


nextmv/nextmv/cli/cloud/version/list.py
```diff
@@ -33,6 +34,10 @@ def list(
     """
     List all versions of a Nextmv Cloud application.
 
+    By default this command paginates the list of tests, which
+    means multiple API calls may be made to retrieve all tests. You may use the
+    --no-pagination option to disable pagination.
+
```
merschformann: I think there is a typo here (should be _versions_ instead of _tests_). 

```suggestion
    By default this command paginates the list of versions, which
    means multiple API calls may be made to retrieve all versions. You may use the
    --no-pagination option to disable pagination.

```

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -549,13 +554,21 @@ def list_runs(self: "Application", status: StatusV2 | None = None) -> list[Run]:
...
         """
 
-        response = self.client.request(
-            method="GET",
-            endpoint=f"{self.endpoint}/runs",
-        )
-
         runs = []
-        for resp_run in response.json().get("runs", []):
+        func = self.client.request if no_pagination else self.client.request_with_pagination
```
merschformann, line 558: Indeed sad that the style of all the others is broken for runs. 😕

nextmv/nextmv/cloud/application/_switchback.py
```diff
@@ -123,10 +123,14 @@ def delete_switchback_test(self: "Application", switchback_test_id: str) -> None
...
-    def list_switchback_tests(self: "Application") -> list[SwitchbackTest]:
+    def list_switchback_tests(self: "Application", no_pagination: bool = False) -> list[SwitchbackTest]:
         """
         List all switchback tests.
 
+        Pagination is enabled by default, but you can disable it with the
+        `no_pagination` argument. With pagination enabled, this function will make
+        multiple API calls if necessary to retrieve all entities.
+
         Returns
```
merschformann, line 139: I think the new parameter is missing in the doc string.

```suggestion
        multiple API calls if necessary to retrieve all entities.

        Parameters
        ----------
        no_pagination : bool, default=False
            Whether to disable pagination when listing entities.

        Returns
```

nextmv/nextmv/cloud/client.py
```diff
@@ -726,6 +954,121 @@ def __set_headers_api_key(self, api_key: str) -> None:
...
+            If both `data` and `payload` are provided.
+            If the `payload` or `data` size exceeds `_MAX_LAMBDA_PAYLOAD_SIZE`.
+            If the maximum number of pages is exceeded.
+            If the API response structure is invalid or unexpected.
+        """
+        query_params = (query_params or {}) | {"pagereturn": "true"}
+        page_count = 0
+
+        while True:
+            if page_count >= max_pages:
```
merschformann, line 1017: The `max_pages` parameter is typed as `max_pages: int | None = 1000`, but if a `None` would get passed it would break here. We should either remove the `None` type hint, or set it to the default if passed. 🤷

nextmv/nextmv/cloud/client.py
```diff
@@ -726,6 +954,121 @@ def __set_headers_api_key(self, api_key: str) -> None:
...
+    def _paginate(
+        self,
+        method: str,
+        endpoint: str,
+        data: Any | None = None,
+        headers: dict[str, str] | None = None,
+        payload: dict[str, Any] | None = None,
+        query_params: dict[str, Any] | None = None,
+        json_configurations: dict[str, Any] | None = None,
+        max_pages: int | None = 1000,
```
merschformann: 1000 pages == 1000 calls to list, right? It feels quite aggressive. Then again, it's not easily override-able, so, maybe better to aim high. 🤷

merschformann (approved):
Thanks for the work! I only had minor comments. I hope they're helpful, no need to block for them though. 😊

nextmv/nextmv/cli/cloud/version/list.py
sebastian-quintero: Nice catch!

nextmv/nextmv/cloud/application/_run.py
sebastian-quintero: LOL, runs is it's own little thing.

nextmv/nextmv/cloud/application/_switchback.py
sebastian-quintero: Yup, good catch.

nextmv/nextmv/cloud/client.py
sebastian-quintero: Yeah good point, I omitted the `| None` signifying that `None` is not ok.

nextmv/nextmv/cloud/client.py
sebastian-quintero: Maybe I should tone it down to 100. Most endpoints cap the entities at `75`, so 1000 * 75 = 75,000 which seems unrealistic, unless we are talking about "real-time" runs. I'll cap it at 500....

nextmv/nextmv/cli/cloud/run/list.py
```diff
@@ -63,11 +66,14 @@ def list(
...
         $ [dim]nextmv cloud run list --app-id hare-app --status queued[/dim]
+
+    - Get the list of runs for an app with ID [magenta]hare-app[/magenta] without pagination.
+        $ [dim]nextmv cloud run list --app-id hare-app --no-pagination[/dim]
     """
 
     cloud_app, _ = build_cloud_app(app_id=app_id, profile=profile)
     in_progress(msg="Listing app runs...")
-    runs = cloud_app.list_runs(status=status)
+    runs = cloud_app.list_runs(status=status, no_pagination=no_pagination)
```
cowanwalls: is there a reason here that we should not or cannot simplify this like the others?

```suggestion
    runs = cloud_app.list_runs(status, no_pagination)
```

nextmv/nextmv/cloud/application/_acceptance.py
```diff
@@ -130,10 +130,19 @@ def delete_acceptance_test(self: "Application", acceptance_test_id: str) -> None
...
         List all acceptance tests.
 
+        Pagination is enabled by default, but you can disable it with the
+        `no_pagination` argument. With pagination enabled, this function will make
+        multiple API calls if necessary to retrieve all entities.
+
+        Parameters
+        ----------
+        no_pagination : bool, default=False
+            Whether to disable pagination when listing entities.
```
cowanwalls: nit: unchanged from entities

```suggestion
            Whether to disable pagination when listing acceptance tests.
```

cowanwalls (approved):

nextmv/nextmv/cli/cloud/run/list.py
sebastian-quintero: No reason, just going with what was already in the code. Looks good.

PR #329, opened 2026-06-12
Enables run comparison in the SDK and CLI, for both `cloud` and `local`

sebastian-quintero:
# Description

Enables run comparison.

- SDK: `cloud.Application.compare_runs()` and `local.Application.compare_runs()` methods were added.
- CLI: `nextmv cloud run compare` and `nextmv local run compare` commands were added.

This feature allows up to 20 runs to be compared, although I don't know how nicely that renders...

By default, tables are printed with the CLI

image

You can use the `--flat` flag to print `json`. The `--flat` naming is not great, but it is meant to be the same flag that is used in `community list`: specifying `--flag` "turns off" table mode... I can also name it something else, no biggie.

image

Specifying the `--output` flag enables `--flat` mode and saves the file as `json`.

### DIFF
--- a/nextmv/nextmv/cli/cloud/run/compare.py
+++ b/nextmv/nextmv/cli/cloud/run/compare.py
@@ -27,8 +27,7 @@ def compare(
         typer.Option(
             "--run-ids",
             "-r",
-            help="List of run IDs to compare (max 20). "
-            "Pass multiple run IDs by repeating the flag, or separating with commas.",
+            help="List of run IDs to compare. Pass multiple run IDs by repeating the flag, or separating with commas.",
             metavar="RUN_IDS",
         ),
     ],
@@ -55,13 +54,11 @@ def compare(
     """
     Compare multiple Nextmv Cloud application runs.
 
-    The number of run IDs provided through the --run-ids flag must be between
-    [magenta]1[/magenta] and [magenta]20[/magenta]. By default this command
-    prints a human-readable table to [magenta]stdout[/magenta]. You may use the
-    --flat option to print the comparison result to [magenta]stdout[/magenta]
-    as [magenta]json[/magenta] instead. When the --output option is used, the
-    --flat flag is automatically activated and the result is saved as
-    [magenta]json[/magenta].
+    By default this command prints a human-readable table to
+    [magenta]stdout[/magenta]. You may use the --flat option to print the
+    comparison result to [magenta]stdout[/magenta] as [magenta]json[/magenta]
+    instead. When the --output option is used, the --flat flag is automatically
+    activated and the result is saved as [magenta]json[/magenta].
 
     [bold][underline]Examples[/underline][/bold]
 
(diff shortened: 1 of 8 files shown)


nextmv/nextmv/cli/cloud/run/compare.py
```diff
@@ -0,0 +1,219 @@
...
+            help="Print the comparison result as [magenta]json[/magenta], instead of a table.",
+        ),
+    ] = False,
+    output: Annotated[
+        str | None,
+        typer.Option(
+            "--output",
+            "-o",
+            help="Saves the comparison result to this location. Activates the --flat option.",
+            metavar="OUTPUT_PATH",
```
merschformann, line 48: I think the json output format is good, but I would also have loved csv for this purpose. 😊

nextmv/nextmv/cli/local/run/compare.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+@app.command()
+def compare(
+    run_ids: Annotated[
+        list[str],
+        typer.Option(
+            "--run-ids",
+            "-r",
+            help="List of run IDs to compare (max 20). "
```
merschformann: I get that we are limiting to 20 for terminal output, but couldn't we also just allow more? Might be useful when just interested in `--output` and the user actively chooses to compare so many, so, fine if they want to maybe. Just a thought. 😊

nextmv/nextmv/run.py
```diff
@@ -2068,3 +2068,164 @@ def logs_text(self) -> str:
...
+    ```python
+    from nextmv import ComparisonValues
+    ```
+
+    Parameters
+    ----------
+    has_differences : bool
+        `True` if the values are not all the same. For example, if
+        we have runs `foo` and `bar` with `execution_class` `fluff` and `tail`,
+        then `has_differences` would be `False`, because `fluff` and `tail` are
```
merschformann: I think this is a type. It states `has_differences` would be `False` when comparing different values `fluff` and `tail`, but it should be `True` since the values **are** different, no? 🤔 

```suggestion
        then `has_differences` would be `True`, because `fluff` and `tail` are
```

merschformann (approved):
Awesome feature, thanks for the work! 🤗 

Just left some thoughts and a typo comment. 😊

nextmv/nextmv/cli/cloud/run/compare.py
sebastian-quintero: One can easily use a terminal tool like `mlr` to translate json to csv, so I think it is no biggie. I think we should get the release going so I'll just leave it as json and can always incorporate other formats if people need/ask for it.

nextmv/nextmv/cli/local/run/compare.py
sebastian-quintero: Ok, let me remove the limit.

nextmv/nextmv/run.py
sebastian-quintero: Nice catch, yep!

PR #338, opened 2026-06-19
Improves formatting of CLI examples and changes an internal file to use internal convention (_)

sebastian-quintero:
# Description

To improve the look of auto-generated CLI reference docs, we need to add some missing spaces. Additionally, marks an internal file with the private Python convention of starting with an underscore. The reason is that the `uv_handler.py` did not contain any public-facing functionality.


nextmv/tests/test_uv_handler.py
merschformann: Small update for full alignment. ☺️ 

```suggestion
Unit tests for the _find_uv_binary() function in nextmv._uv_handler.
```

merschformann (approved):
🎉

PR #341, opened 2026-07-07
Fix `BaseModel` serialization and scenario test creation

sebastian-quintero:
# Description

There was an issue when the refactor for `app.yaml` took place. The refactor was done to automatically configure paths, metrics, and more stuff based on the `app.yaml` being present at the root. Quietly, serialization for `BaseModel` stopped working, meaning that when we do stuff like `nextmv.write(input_set)`, where `input_set` is one of our classes, the function didn't write anything. This should fix it now.

This PR also fixes scenario test creation. When using the CLI, we weren't correctly parsing the JSON scenarios. It worked with the SDK because we were using classes directly. The fix is to use `BaseModel` class to deserialize JSON scenarios correctly. Also, the help message for `nextmv cloud scenario create` was off.

### DIFF
--- a/nextmv/nextmv/__about__.py
+++ b/nextmv/nextmv/__about__.py
@@ -1,1 +1,1 @@
-__version__ = "v1.9.6.dev0"
+__version__ = "v1.9.6.dev1"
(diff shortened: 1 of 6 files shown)


merschformann (approved):
Thanks for the fix. There is a failing test, fyi. ❤️

PR #344, opened 2026-07-27
Fix: cache deletion, run comparison, help msg for queuing, input sets for acceptance, input set update. Adds: `--yes-queuing` option, ensemble result to run result

sebastian-quintero:
# Description

This PR addresses several issues that were found during the update of the docs.

## Features
- Adds the `--yes-queuing` flag to turn queuing back on. When an instance is updated to turn off queuing, there was no way to turn it back on.
- Adds the `.ensemble` result to a run result that is an `ensemble` run.

## Fixes
- Cache deletion logic. The cache was being deleted before asking the user for confirmation.
- Run comparison. Runs with nested metrics under a `metrics` key could not be compared. This special case was addressed to flatten the metrics.
- Help message for queuing. Given that the `--yes-queuing` option is being added to the CLI, the help message was adjusted.
- Input sets for acceptance tests. When creating a new acceptance test, an input set created from runs could not be used. This was fixed so that input sets created from either runs or managed inputs can be used for acceptance tests.
- Input set updating. When attempting to update an input set's name/description, the inputs would get cleared out. This was fixes so that we don't have to always give updated inputs to the input set when updating it. What is intended to be patch behavior is now rectified.

### DIFF
--- a/nextmv/nextmv/cli/cache/delete.py
+++ b/nextmv/nextmv/cli/cache/delete.py
@@ -4,8 +4,8 @@ This module defines the cache delete command for the Nextmv CLI.
 
 import typer
 
-from nextmv.cache import clear_cache, format_bytes
-from nextmv.cli.message import confirmation, info, success
+from nextmv.cache import clear_cache, format_bytes, get_cache
+from nextmv.cli.message import confirmation, in_progress, info, success
 from nextmv.cli.options import DebugOption, YesOption
 
 # Set up subcommand application.
@@ -31,11 +31,13 @@ def delete(yes: YesOption = False, _: DebugOption = False) -> None:
         $ [dim]nextmv cache delete --yes[/dim]
     """
 
-    deps, bytes_deleted = clear_cache()
-    if deps == 0 and bytes_deleted == 0:
+    cache_info = get_cache()
+    deps, total_size = cache_info["num_dependencies"], cache_info["total_size"]
+    if deps == 0 and total_size == "0 B":
         info("Nothing to delete: cache is already empty.")
         return
 
+    in_progress(f"Deleting [magenta]{deps}[/magenta] dependencies ([magenta]{total_size}[/magenta]) from the cache...")
     if not yes:
         confirm = confirmation(
             "Are you sure you want to delete the Nextmv cache? This action cannot be undone.",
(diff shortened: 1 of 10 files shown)


nextmv/nextmv/cli/cache/delete.py
```diff
@@ -31,11 +31,13 @@ def delete(yes: YesOption = False, _: DebugOption = False) -> None:
...
 
-    deps, bytes_deleted = clear_cache()
-    if deps == 0 and bytes_deleted == 0:
+    cache_info = get_cache()
+    deps, total_size = cache_info["num_dependencies"], cache_info["total_size"]
+    if deps == 0 and total_size == "0 B":
         info("Nothing to delete: cache is already empty.")
         return
 
+    in_progress(f"Deleting [magenta]{deps}[/magenta] dependencies ([magenta]{total_size}[/magenta]) from the cache...")
```
merschformann, line 40: ❤️

nextmv/nextmv/cloud/application/_run.py
```diff
@@ -459,6 +460,11 @@ def compare_runs(self: "Application", run_ids: Iterable[str]) -> RunComparison:
 
             # Gather the metrics.
             metrics = metadata.metrics
+
+            # If metrics come nested, we extract them.
+            if len(metrics.keys()) == 1 and list(metrics.keys())[0] == METRICS_KEY:
+                metrics = metadata.metrics[METRICS_KEY]
```
merschformann: According to `Metadata` definition metrics can be `None` (the type is `dict[str, Any] | None`). So, this would now invoke `metrics.keys()` **before** the non-None check a little lower. We should either guard it with an additional check (see suggestion), or, move it inside of the `if metrics:` check block below.

```suggestion
            if metrics is not None and len(metrics.keys()) == 1 and list(metrics.keys())[0] == METRICS_KEY:
                metrics = metadata.metrics[METRICS_KEY]
```

merschformann (approved):
Nice, just had a small improvement suggestion. ❤️ 

Also, tests are failing. :sweat_smile:

cowanwalls (commented):
I'm digging in to the backend code to double check the source of the error we're seeing in the integration tests (it is consistent across 3.10 and 3.13 and is an HTTP 400 with message `Run type does not support detail information.`). to see if I can identify what change here caused it. I'll request changes with the code in question marked once I have done the digging.

cowanwalls (changes requested):
[These lines](https://github.com/nextmv-io/nextmv-py/blob/dfc0931a2dbad86aac06893f2635aa8d43a8759b/nextmv/tests/integration/cloud/test_integration_cloud.py#L900-L922) cause us to run an ensemble with only one run. This trips a bug in the backend that causes the ensemble run details to become inaccessible. The quick fix for this is to define a `repetitions` value for the ensembles in the test code to a value >=1 while we work on bugfixes for the backend.

nextmv/nextmv/cloud/application/_run.py
sebastian-quintero: Thank you! I moved the check inside the `if` clause below.

cowanwalls (approved):

PR #348, opened 2026-07-28
Fix usage of write function when output is a simple dict

sebastian-quintero:
# Description

Another regression. We lost the ability to write simple dicts with the `nextmv.write` function. This should hopefully be resolved now.


merschformann (approved):
Thank you for the fix! 🤗

PR #353, opened 2026-08-14
Fix down a regression with Typer and pin dependencies

sebastian-quintero:
# Description

Typer `v0.27.0` silently broke us by changing the way an option is parsed. When using `typer.Option`, _not_ providing an explicit option name, but providing a `metavar`, Typer automatically used the metavar to name the option. In practice, what this means is that some options were being shown in all caps, like `--INSTANCE-ID` for the `nextmv cloud run create` command.

image

This means that trying to execute `nextmv cloud run create --instance-id foo` would fail because it would not recognize the option being given, it wanted `--INSTANCE-ID`.

The fix consists of two remediations:

1. Tracked down all places where `typer.Option` was being used to declare an option and the explicit long name was not provided. All long names are now provided. Before, we only provided long names when we wanted to specify a shorthand too. Now, a long name is provided even if a shorthand is not available, like with `--instance-id` in the `nextmv cloud run create` command.
2. Pinned down all the versions of the dependencies. When developing locally, an old version of Typer was used so the options were fine. When using `uv` to install the CLI, because Typer was `>=`, it automatically fetched the latest version of `typer`, and thus the bug was present. To avoid this, all dependencies were pinned to the floor specified in the `pyproject.toml` with a few exceptions. I'll note those exceptions with comments on the files directly.

Please note that I also pinned deps for  `nextmv-gurobipy` and `nextmv-scikit-learn` for consistency. One last thing is that I also cleaned up the `dev` dependency groups. The `dev` group should be used for dependencies used in the development process that are not part of the package. A lot of the main dependencies were duped in the dev group, which is redundant. Made sure that only true dev dependencies are in the dev groups.


nextmv-gurobipy/pyproject.toml
```diff
@@ -13,11 +13,11 @@ classifiers = [
...
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13"
+    "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "gurobipy>=13.0.1",
-    "nextmv>=0.34.1"
+    "gurobipy==13.0.1",
+    "nextmv>=1.10.1",
```
sebastian-quintero, line 20: We are an exception for pinning, to avoid having to open subsequent PRs after every release that update this pinned version. We are leaving ourselves unpinned.

nextmv-scikit-learn/pyproject.toml
```diff
@@ -16,8 +16,8 @@ classifiers = [
     "Programming Language :: Python :: 3.13"
 ]
 dependencies = [
-    "scikit-learn>=1.6.1",
-    "nextmv>=0.34.1"
+    "scikit-learn==1.6.1",
+    "nextmv>=1.10.1",
```
sebastian-quintero, line 20: Same here: we leave ourselves unpinned.

nextmv/pyproject.toml
```diff
@@ -13,21 +13,20 @@ classifiers = [
...
-    "pyyaml>=6.0.1",
-    "typer>=0.20.1",
-    "pip>=26.0",
-    "folium>=0.20.0",
-    "plotly>=6.0.1",
-    "questionary>=2.1.1",
-    "truststore>=0.10.0",
-    "uv>=0.4.0",
-    "tomli>=2.4.1 ; python_full_version < '3.11'",
+    "pydantic==2.11.0",
```
sebastian-quintero: This is an exception. We don't use the same floor as we had (`2.5.2`) because it is incompatible with `mcp`. I bumped the floor a little bit for that reason.

nextmv/pyproject.toml
```diff
@@ -13,21 +13,20 @@ classifiers = [
...
+    "urllib3==2.1.0",
+    "requests==2.31.0",
+    "pyyaml==6.0.1",
+    "typer==0.27.1",
```
sebastian-quintero: Even though they broke us, went ahead and pinned to the latest version... Now we work with the latest version.

nextmv/pyproject.toml
```diff
@@ -63,34 +62,20 @@ nextmv = "nextmv.cli.main:main"
 # Dependencies that are optional for the end user to install.
 [project.optional-dependencies]
 mcp = [
-    "mcp[cli]>=1.0",
+    "mcp[cli]==1.2.0",
```
sebastian-quintero: This is another exception. Turns out this was wrong. The version to introduce the `cli` extra for `mcp` was `1.2.0`, not `1.0.0`. I went ahead and used that floor.

nextmv/pyproject.toml
```diff
@@ -13,21 +13,20 @@ classifiers = [
...
-    "Programming Language :: Python :: 3.13"
+    "Programming Language :: Python :: 3.13",
 ]
 dependencies = [
-    "pydantic>=2.5.2",
-    "urllib3>=2.1.0",
-    "requests>=2.31.0",
-    "pyyaml>=6.0.1",
-    "typer>=0.20.1",
-    "pip>=26.0",
```
sebastian-quintero, line 24: @merschformann I need your explicit attention here. I checked that `pip` was not used anywhere, given we changed to `uv`. Can you confirm that deleting `pip` as a dependency is ok?

nextmv/pyproject.toml
```diff
@@ -13,21 +13,20 @@ classifiers = [
...
+    "pydantic==2.11.0",
+    "urllib3==2.1.0",
+    "requests==2.31.0",
+    "pyyaml==6.0.1",
+    "typer==0.27.1",
+    "folium==0.20.0",
+    "plotly==6.0.1",
+    "questionary==2.1.1",
+    "truststore==0.10.0",
+    "uv==0.9.29",
```
sebastian-quintero: `uv` is another exception. I had to bump it because `0.4.0` does not support `manylinux_2_34`. I matched this version to the one we are already using in CI/CD.

nextmv-scikit-learn/pyproject.toml
merschformann: I think this makes a lot of sense, to not constrain our users too much. ❤️

nextmv/pyproject.toml
merschformann: Deleting pip should be ok.

nextmv/pyproject.toml
```diff
@@ -13,21 +13,20 @@ classifiers = [
...
-    "pip>=26.0",
-    "folium>=0.20.0",
-    "plotly>=6.0.1",
-    "questionary>=2.1.1",
-    "truststore>=0.10.0",
-    "uv>=0.4.0",
-    "tomli>=2.4.1 ; python_full_version < '3.11'",
+    "pydantic==2.11.0",
+    "urllib3==2.1.0",
+    "requests==2.31.0",
```
merschformann: I am not sure whether fixing all deps here is a good choice. We will start dictating all the versions of these packages for our users who just use `nextmv` to interact with backend for example or for modeling. They will not be able to use newer versions for their projects due to our strict versions of these deps being dictated. For CLI I would be fine with this, but for a lib package I find this way too invasive. 

I might be missing something though. Let me know. 😊

merschformann (commented):
Good find with the param names. I am not sure whether I follow with pinning all deps from `nextmv` package. That feels quite invasive for a lib package. Maybe I am missing a point here though.

nextmv/pyproject.toml
sebastian-quintero: @merschformann you are correct. I was too hung up on not having anyone break us anymore but forgot about the downstream user... I reverted the changes back to `>=`, while keeping the updated dependency versions where applicable.

One difference is the `dev` group. For reproducibility and CI/CD environments, _that_ group makes sense to have dependency pinning with `==`. But the normal dependencies I reverted to `>=`. Thank you for this callout!

sebastian-quintero:
@merschformann I processed your comments, which was mainly reverting to `>=`. I kept the updated floors on the dependencies I mentioned, like Pydantic and mcp. For the `dev` group, I am still using `==` for pinning given that it makes sense from a reproducibility standpoint.

nextmv/pyproject.toml
```diff
@@ -13,21 +13,21 @@ classifiers = [
...
-    "pip>=26.0",
+    "typer>=0.27.1",
     "folium>=0.20.0",
     "plotly>=6.0.1",
     "questionary>=2.1.1",
     "truststore>=0.10.0",
-    "uv>=0.4.0",
+    "uv>=0.9.29",
     "tomli>=2.4.1 ; python_full_version < '3.11'",
+    "mcp[cli,mcp]>=1.2.0",
```
merschformann: Why is this coming in here? This does not feel related to a typer fix. 🤔 

Also, the `mcp` extra looks weird, shouldn't it be `mcp[cli]` instead?

I would like to avoid additional dependencies here if they are not a must. If they are, I would like to understand what changed (as it was not part of this PR motivation it feels). ☺️

merschformann (commented):
Left a comment about the mcp dependencies, all fine otherwise. Thanks for the work! ❤️

nextmv/pyproject.toml
sebastian-quintero: FOR SOME REASON this crept into the main dependencies, and with some big typos looks like. Thank you for reviewing, this would have been weird otherwise!

sebastian-quintero:
@merschformann Third time's the charm. That MCP thing crept in for some unknown reason. Now the deps for `nextmv` look as expected. Only the deps I mentioned in various comments changed: `typer`, `pydantic`, and `uv`. The `mcp[cli]` got updated too, for the reasons I already explained. The Zizmor thing is failing, and I have never seen that check before... Did you enable it?

merschformann (approved):
Thanks for the work! Looks all good! 🤗

merschformann:
> @merschformann Third time's the charm. That MCP thing crept in for some unknown reason. Now the deps for `nextmv` look as expected. Only the deps I mentioned in various comments changed: `typer`, `pydantic`, and `uv`. The `mcp[cli]` got updated too, for the reasons I already explained. The Zizmor thing is failing, and I have never seen that check before... Did you enable it?

The zizmor one we added a while ago for more security checks being in place. It only failed yesterday due to Github being flaky, I think. I nudged it. :relaxed:

PR #356, opened 2026-08-18
Bundle local path and direct dependencies on push

merschformann:
# Description

The uv-based dependency bundler only kept `name==version` lockfile lines, so local path deps (e.g. `../../`) and direct URL/VCS references were silently dropped from the pushed app artifact — transitive deps got bundled, but the referenced package itself did not.
This was a small, albeit exotic, regression from previous pip non-cached implementation.

## Changes

- Parse non-pinned lockfile lines (local paths, direct URL/VCS) instead of discarding them.
- Build and bundle those requirements from source (fresh each push, uncached).
- Add unit tests, including a regression test for the dropped local path dep.

🤖 Generated with help of [Claude Code](https://claude.com/claude-code)


dirkschumacher (approved):

nextmv/nextmv/cloud/package.py
```diff
@@ -991,6 +1172,38 @@ def _log_pkg_cache_status(
...
+    rich_print : bool
+        Whether to use rich printing for verbose logs.
+    direct_count : int
+        The number of local/direct requirements being built from source.
+    """
+
+    if not verbose or direct_count <= 0:
+        return
+
+    dep_word = "requirement" if direct_count == 1 else "requirements"
```
cowanwalls: Alternatively we could inline `requirement(s)`, if we want the string to always be consistent aside from the (indirectly) user defined number. This construct communicates this same information, but is less formal. The ternary operation here has negligible cost of any kind, so it is more a stylistic consideration than anything else.

cowanwalls (approved):

sebastian-quintero (approved):
Lgtm

nextmv/nextmv/cloud/package.py
merschformann: Yeah, I simplified it a bit. Just for a bit more stable / reliable CLI tool output. No hard opinions here either. 😅

cowanwalls (approved):

PR #363, opened 2026-09-17
Pin `mcp` extra to `<2` and surface import errors

sebastian-quintero:
# Description

An un-gated `mcp>=` as an extra dependency resolved to a `v2` that broke the CLI MCP server. We are pinning the dependency to `v1x`.


merschformann (approved):
❤️

